PR #21, opened 2021-09-19
Optionally embed metadata into video files

nifoc:
This adds basic `postprocessors` support and adds a setting to optionally enable `FFmpegMetadata`.

It's the same as passing `--add-metadata` and `--add-chapters` via the `yt-dlp` CLI.

---

The feature itself seems to work as expected, but I'm not sure if I have really added the new setting in all the right places.
For example, the *current value* text in `settings.html` was empty before I actually set it to something. Is there a way to actually set default values for new settings?

### DIFF
--- a/tubearchivist/home/config.json
+++ b/tubearchivist/home/config.json
@@ -14,7 +14,8 @@
         "limit_count": 5,
         "limit_speed": false,
         "sleep_interval": 3,
-        "format": false
+        "format": false,
+        "add_metadata": false
     },
     "application": {
         "cache_dir": "/cache",
--- a/tubearchivist/home/src/download.py
+++ b/tubearchivist/home/src/download.py
@@ -454,6 +454,19 @@ class VideoDownloader:
         external = False
         if external:
             obs['external_downloader'] = 'aria2c'
+
+
+        postprocessors = []
+
+        if self.config['downloads']['add_metadata']:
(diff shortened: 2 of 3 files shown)


bbilly1:
Nice, very useful. Thank you for your contribution. 

As I have it setup now is that the file `config.json` only gets read if Redis doesn't have it already stored, so basically only on new installs or as a fallback if Redis doesn't have a volume set for some reason. Then once it's read, it will take precedence over the default and every config changes only life in Redis. That's why you'll only see it on settings.html once you have it actually changed. 

So yes, that's not ideal and we need something in the [AppConfig](https://github.com/bbilly1/tubearchivist/blob/testing/tubearchivist/home/src/config.py) class that can pick up new default values without touching the already changed variables. I'm sure this will be useful again later down the road...

Let me take a look at that and then we can test your pull request again.

bbilly1:
OK, as far I have been able to test it [4b9acd4](https://github.com/bbilly1/tubearchivist/commit/4b9acd4c513d036c2767f8ef801ffea75cee5bb9) is now reading new config.json values if they are not set yet. So I added it as part of the startup routine. I did also some refactoring in the AppConfig class at that opportunity...

Do you want to give it a test?

nifoc:
Seems to work for me, too! 🎉

PR #86, opened 2021-11-08
Update settings.html

pairofcrocs:
Fixed the example as you wouldn't put KB/s at the end, reworded the explanation, and added "KB/s" at the end of the "Current throttled rate limit:"


bbilly1:
Yes, good catch, that disappeared with the new validated forms... Actually looking at it, it's also missing for *download speed limit* as well. Do you want to add it there as well? That's KB/s too. And then I think it's *amount* in stead of *ammount*, or at least codespell thinks it is. :-)
Ah and if you don't mind, please make the pull request against the testing branch instead of master for easy merging.

PR #137, opened 2022-01-10
Added a button to download a video to the local device.

n8detar:
This is my first time doing this so please let me know if I've done something incorrectly.

I added a download button on the video page to the left of the `Delete Video` button called `Download Video`. This name could be confused as a button to have their server download the video, but I couldn't think of a better name for this button.

Currently, if you want to download a video from Tube Archivist to the local device you have to wait until the page and the video finishes loading then select the menu button, and then select Download. It can take a while for the video to finish loading while on mobile or if you are accessing Tube Archivist on a slow connection. Having a button to immediately download the video to the local device is more convenient and gives the user a way to watch the video if their connection is slow or if they are planning on going somewhere without any internet.

### DIFF
--- a/tubearchivist/home/templates/home/video.html
+++ b/tubearchivist/home/templates/home/video.html
@@ -45,7 +45,7 @@
                 {% else %}
                     <p>Youtube: Deactivated</p>
                 {% endif %}
-                <a download="" href="/media/{{ video.media_url }}"><button id="download-item">Download Video</button></a>
+                <a download="" href="/media/{{ video.media_url }}"><button id="download-item">Download File</button></a>
                 <button onclick="deleteConfirm()" id="delete-item">Delete Video</button>
                 <div class="delete-confirm" id="delete-button">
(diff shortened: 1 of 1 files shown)


insuusvenerati:
Can you post a screenshot of it? :)

n8detar:
> Can you post a screenshot of it? :)

Screenshot from 2022-01-09 16-39-02

n8detar:
Just noticed the theme was slightly off. I moved the <button> tag into the <a> tag which fixed it.
Normal

bbilly1:
How about calling it `Download File`? That could be a good distinction between *downloading video from YouTube to TubeArchivist* and *downloading file from TubeArchivist to your device*. Then a confirmation box from the browser will pop up to confirm or cancel anyways, so I think at least at that point it will be clear whats happening.

n8detar:
> How about calling it `Download File`? That could be a good distinction between _downloading video from YouTube to TubeArchivist_ and _downloading file from TubeArchivist to your device_. Then a confirmation box from the browser will pop up to confirm or cancel anyways, so I think at least at that point it will be clear whats happening.

`Download File` makes sense and is shorter than something like `Download to Device`.

PR #140, opened 2022-01-10
Basic Google Cast Support

n8detar:
After some quick research, I was able to put together very basic cast support. It allows you to start and stop casting the video and that is it. I was only able to make this by using a lot of code from Google's docs and sample code so the implementation may not be the best, but it works and could act as a placeholder until proper support is implemented. I couldn't make sense of any of the more complicated stuff like pause/play, volume controls, or making a proper cast button so this is the extent of what I am able to implement.

Cast Buttons

### DIFF
--- a/tubearchivist/home/config.json
+++ b/tubearchivist/home/config.json
@@ -31,7 +31,8 @@
         "cache_dir": "/cache",
         "videos": "/youtube",
         "file_template": "%(id)s_%(title)s.mp4",
-        "colors": "dark"
+        "colors": "dark",
+        "enable_cast": false
     },
     "scheduler": {
         "update_subscribed": false,
--- a/tubearchivist/home/forms.py
+++ b/tubearchivist/home/forms.py
@@ -62,6 +62,12 @@ class ApplicationSettingsForm(forms.Form):
         ("1", "enable ryd integration"),
     ]
 
+    CAST_CHOICES = [
+        ("", "-- change Cast integration --"),
+        ("0", "disable Cast"),
+        ("1", "enable Cast"),
+    ]
+
     subscriptions_channel_size = forms.IntegerField(required=False)
     downloads_limit_count = forms.IntegerField(required=False)
     downloads_limit_speed = forms.IntegerField(required=False)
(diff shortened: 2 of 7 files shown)


bbilly1:
Very nice! That's a very good start. I think we'll need an option for that on the settings page to enable it, so the buttons won't show if somebody doesn't use it. 

Has also the downside that it pulls in a third party JS library from google... But I think that's the price you pay for convenience. :-) But loading the library could then also be dynamic, so that this only happens if you have cast enabled first.

n8detar:
> Very nice! That's a very good start. I think we'll need an option for that on the settings page to enable it, so the buttons won't show if somebody doesn't use it.
> 
> Has also the downside that it pulls in a third party JS library from google... But I think that's the price you pay for convenience. :-) But loading the library could then also be dynamic, so that this only happens if you have cast enabled first.

I agree having it optional would be a good idea as I'm sure there will be users who wouldn't be too happy about having a library from Google. The JS was just a quick and easy way to implement it.

I found a python library called PyChromecast that Home Assistant uses. I know TA uses python on the backend but I don't know how realistic it would be to use that library as an alternative. It might be an option for a more permanent implementation.

n8detar:
> Very nice! That's a very good start. I think we'll need an option for that on the settings page to enable it, so the buttons won't show if somebody doesn't use it.

From what I saw for code controlled by a setting like the theme color I would do something along the lines of this right?
```
{% if cast %}
    <button onclick="startCast()">Start Cast</button>
    <button onclick="stopCast()">Stop Cast</button>
{% endif %}
```
Same thing for the JS code if the settings variable was for example cast.
```
    {% if cast %}
        <script src="https://www.gstatic.com/cv/js/sender/v1/cast_sender.js?loadCastFramework=1"></script>
        <script>
            window['__onGCastApiAvailable'] = function(isAvailable) {
              if (isAvailable) {
                initializeCastApi();
              }
            };
        </script>
        <script type="text/javascript" src="{% static 'cast-videos.js' %}"></script>
    {% endif %}
```

bbilly1:
> From what I saw for code controlled by a setting like the theme color I would do something along the lines of this right?

That's it, you got it! Feel free to add the commit for both of your snippets, then I'll implement the settings page to make it configurable.

n8detar:
> That's it, you got it! Feel free to add the commit for both of your snippets, then I'll implement the settings page to make it configurable.

Ok, I added it. I just realized that that syntax is the same as what Home Assistant uses for templating, I was wondering why it seemed familiar.

bbilly1:
OK, this is now published with the `unstable` build on docker. If you want to take a look and let me know how this works over on Discord.

PR #141, opened 2022-01-11
Added cast integration requirements.

n8detar:
HTTPS and a supported browser are required for this integration.

Cast Settings


bbilly1:
So, if I'm hearing you right, that untested span is no longer required then! Do you want to do the honers? :-)

bbilly1:
Just a heads up, you should treat your API token like your password, this will give full access to everything. Maybe this should be hidden by default, like the video descriptions...

n8detar:
> Just a heads up, you should treat your API token like your password, this will give full access to everything. Maybe this should be hidden by default, like the video descriptions...

This is on my dev server which isn't exposed to anything, but thanks for the reminder. I forgot that that was added.

PR #143, opened 2022-01-12
Improved Cast Support

n8detar:
Improved implementation of cast integration. Added proper cast button with basic theming and redid and simplified some of the backend JS. The title of the video is also sent to the cast device.

Full
Cast Button OFF/ON
Cast Button offCast Button ON

### DIFF
--- a/tubearchivist/home/templates/home/base.html
+++ b/tubearchivist/home/templates/home/base.html
@@ -30,7 +30,7 @@
     <script type="text/javascript" src="{% static 'script.js' %}"></script>
     {% if cast %}
         <script type="text/javascript" src="https://www.gstatic.com/cv/js/sender/v1/cast_sender.js?loadCastFramework=1"></script>
-        <script type="text/javascript" src="{% static 'cast-videos.js' %}"></script>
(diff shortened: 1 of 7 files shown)


bbilly1:
Oh nice, that looks good. I'll double check things a bit later. As far as I can see that will replace PR #141, right?

n8detar:
I don't think the `cast` integration variable is being set on the home page. I tested without the setting condition and everything is working on both the videos page and the home page. So once the variable issue is cleared up this should be good to go.

n8detar:
The latest commit makes it so the casted video starts at the same time position as it was in the browser. So if you start playing the video in the browser but then decide to cast it you don't need to skip to get back to the position you were at. With that added and the checks passing I think the integration is at a good place. I've done a lot of testing so I don't think there are any bugs left but other people should test it out in case I've missed something, so I won't add any more commits so that this can maybe finally be merged lol.

bbilly1:
I've made a new unstable build. As far as I can tell, the cast variable should now be available throughout the views containing videos so the JS library will load if needed.

PR #150, opened 2022-01-14
Invert hide videos button behavior to better reflect current setting

GigaFyde:
Currently when set to hide already watched videos. it'll look like this: 

Displays as off, but is actually set to on

As you can see, the label says off. but actually its currently set to hide already watched videos.
And here with it set to the off position.

Displays as on, but is actually set to off

 This PR aims to invert this behavior. by swapping the two around.

With the new "on" setting being:
Correctly displays as on

And off being as follows:
Correctly displays as off


bbilly1:
There is a new unstable build containing your changes. Thank you for your help!

PR #151, opened 2022-01-15
Get Video Player Data Using New API

n8detar:
The `videoPlayer()` function now gets it's data from the API rather then the HTML. It still pulls the video id from the button. 3 functions were also added `getVideoPlayerData()`, `getVideoData()`, and `apiRequest()`. `apiRequest()` makes an api request when passed an endpoint (ex. `/api/video/VIDEO_ID/player/`) and a method (either "GET" or "POST") and returns the results in JSON. `getVideoPlayerData()` returns video player data in JSON when it is given a video ID (Makes a call to `apiRequest()` ). `getVideoData()` isn't used right now, it's just another example and returns video data in JSON when it is given a video ID (It also makes a call to `apiRequest()`.


bbilly1:
Very nice! Thank you for looking into it. I see you have figured out how to grab the session id from the cookie. :-) 

Now it's time to reap the rewards and use the api everywhere there are inplace play buttons: channel_id.html, playlist_id.html and also on the search page. The video item there is generated with javascript markup [here](https://github.com/bbilly1/tubearchivist/blob/26bc2d2af45007f66c7017b4375a8c1c0775a6b8/tubearchivist/static/script.js#L445). That can now be simplified too. 

Then I think I understand what you are doing, you can remove some of the comments and maybe keep a short line on top of the function to describe what it's doing. 

Thank you very much for all your help!

n8detar:
`createPlayer()` now uses the markup variable and the API. `createVideo()`, `createChannel()`, and `createPlaylist()` have also been switched to the API and extra metadata in `channel_id.html`, `playlist_id.html` has been removed. Other than the video's publish date, last channel refresh, and last playlist refresh the metadata is now being pulled from the API. The API reports dates in different formats like for a video's publish date it is in `YYYY-MM-DD`, but for the last refresh it is a number like `1642563554` which I thought was milliseconds but it didn't seem to convert correctly so I left the dates the way they were originally since on the front end they are formatted like `09 Jan, 2022`.

n8detar:
An API endpoint with integrations status would be helpful. It would make it easier to check if the cast button should be displayed. Right now I'm just checking if the script is added to the page or not. It would also help with stars and downvotes as well for the returnyoutubedislike integration or any future integrations.

n8detar:
If you add an endpoint for video playback position, maybe under player where the watched variable is, I can implement saving playback position. The endpoint would just need to accept a number as its value.

n8detar:
> Other than the video's publish date, last channel refresh, and last playlist refresh the metadata is now being pulled from the API.

I figured out the dates and switched all the functions over to the API.

bbilly1:
Sorry, I've let you hanging on this one.

So what stands out immediately is that, the API endpoints really need to return better formatted data:
- Date formatting shouldn't have to be done in the frontend, the API should already return the correct format for everything.
- Creating the player should require one, and only one API call, which should return all relevant data.

I like that the inplace player has more information now, but I think it's too much. I don't think that this needs to be able to replace the dedicated video page. It should all fit into 100vh with space to breath so you get the "theater" feeling. That's why I tried to visually distinct it with the *highlight-bg* color.

So I don't think we'll have enough space for the three description divs there, so for the channel, I think just the name with link will have to suffice, for the middle div with the published, last refresh, youtube link and buttons, I think that's fine to only have on the video detail page, published date is also in the grid... then maybe we'll be able to compact the views, likes and dislikes? I think that would be very nice to have there.

maybe something like:
\<thumbs-up-img>: 10k | \<thumbs-down-img> 500 | views: 1.2m

Then maybe we could also have playlist links there? That would be very useful!

n8detar:
> Creating the player should require one, and only one API call, which should return all relevant data.

For your second point, I think just merging the `/video/videoId/player/` endpoint into the `/video/videoId` one should be fine, I don't think there needs to be a second endpoint since the player data really is just more data for the video.

n8detar:
> Then maybe we could also have playlist links there? That would be very useful!

The playlist variable in the video endpoint is an array correct, should it just link to the first playlist or maybe first subscribed? I'm not sure how tubearchivist handles a video if it is in two playlists.

bbilly1:
> I don't think there needs to be a second endpoint since the player data really is just more data for the video.

Yes, I get your point. Not too hard to grab the needed fields from the JSON even if it returns some unneeded keys.

> I'm not sure how tubearchivist handles a video if it is in two playlists.

That's not very elegant now, as there is no path to "remember" in which playlist you are now for example with an url query string... I think that will require a bit more work to make this better first. Let's put a pin into that and revisit it later.

n8detar:
> maybe something like: \<thumbs-up-img\>: 10k | \<thumbs-down-img\> 500 | views: 1.2m

An icon for views might be a good idea, maybe an eye, to better match the likes and dislikes.

bbilly1:
> An icon for views might be a good idea, maybe an eye, to better match the likes and dislikes.

Excellent idea! Let me see, what I can do.

n8detar:
New Player
Ignore the cast button color, I forced it to be visible since I'm not home right now.  I'm thinking the views/likes/etc. should have more space between them, but what do you think? I still need to test on smaller screens.

bbilly1:
very nice! maybe we don't even need the rating? just have the other three values? then let me also see if I can better center the thumbs up icon, so it looks even when it's rotated and next to each other like that...  and yes, could need a few pixels between the fields.

n8detar:
I noticed that the old player had the channel name above the title so I switch to that which helped a bit with the mobile view of the stats but will need some more work. I'll do some cleaning up then push the changes.

n8detar:
Added first playlist the video is in that the user is subbed to
New Player WIth Playlist

n8detar:
Currently mobile it is a bit crowded but everything more or less formats correctly even on a very small screen, ex iPhone SE:
New Player Very Small

On medium-sized screens and lower it might be better to just have the views/likes left aligned but IDK how to do that. Once the title starts to overlap with the stats it doesn't look as good.
New Player Medium

bbilly1:
OK, I've played around a bit with the position of the stats, I think if we put it next to the watched checkbox, this will a one solution to fit all screens. Or what do you think?

This will allow future expansion to add some playlist shortcuts below there.

n8detar:
I think it looks great. I just removed the dash in front of the playlist and reordered the cast button and the stats.
Reorder

n8detar:
@bbilly1 Other than maybe using the API to check for integration status for casting and the downvotes, the player is probably good to go. The API isn't necessary at the moment, but I think that it is something that would be useful for future features.
So unless there is anything else that needs to be cleaned up it would probably be good to have this tested and get feedback from users, they might come up with some changes that we didn't think of.

Additionally, if an endpoint is added that accepts input for video progress I can pretty easily add a save progress feature. It would just need to accept a number to represent the number of seconds into the video the user has watched.

PR #167, opened 2022-02-10
When there is no videos, add hint to start a search and download

SteVwonder:
Per a discord discussion.

HTML+CSS styling are not my strong suit, and I’m editing on an iPad with (I think) limited dev console capabilities.  So I’m totally open to suggestions on possible better formatting.

### DIFF
--- a/tubearchivist/home/templates/home/channel_id.html
+++ b/tubearchivist/home/templates/home/channel_id.html
@@ -133,6 +133,7 @@
             {% endfor %}
         {% else %}
             <h2>No videos found...</h2>
+            <p>Try going to the <a href="{% url 'downloads' %}">downloads page</a> to start the scan and download tasks.</p>
         {% endif %}
     </div>
 </div>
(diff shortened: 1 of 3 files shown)


bbilly1:
Nice! I think that makes a lot of sense. 

Django has this convenient feature, that it can create links based on the view name like used here:
https://github.com/bbilly1/tubearchivist/blob/5b37bd059cb35dbbc64260ef2ea49f010604fbd2/tubearchivist/home/templates/home/base.html#L60

This has the advantage that all links to the same page are referred from one place in urls.py here:
https://github.com/bbilly1/tubearchivist/blob/5b37bd059cb35dbbc64260ef2ea49f010604fbd2/tubearchivist/home/urls.py#L34

This will help in feature maintainability. 

Link to the [documentation](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#url).

Thanks for your help!

PR #169, opened 2022-02-11
Fix cast support to handle new video tag format.

n8detar:
Move `video-item` id to source tag.

### DIFF
--- a/tubearchivist/home/templates/home/video.html
+++ b/tubearchivist/home/templates/home/video.html
@@ -4,8 +4,8 @@
 {% load humanize %}
 <div class="video-main">
     <video poster="/cache/{{ video.vid_thumb_url }}" controls preload="false" width="100%" playsinline 
-    id="video-item" ontimeupdate="onVideoProgress('{{ video.youtube_id }}')" onloadedmetadata="setVideoProgress(0)">
-        <source src="/media/{{ video.media_url }}" type="video/mp4">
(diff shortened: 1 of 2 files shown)


bbilly1:
It's possible that this has nothing to do with your implementation, as far as I can tell, this looks fine. I've scrolled through the SDK documentation and most likely its [CORS](https://developers.google.com/cast/docs/web_sender/advanced#cors_requirements) that's messing with us here. 

For now I'd suggest to revert the last commit so the media streaming is working. Then we can look into this further for a future release. 

So I can merge this, and we'll have cast support in v0.1.1! :-)

bbilly1:
Ah, can you also add a note in the wiki? Short description on how to set it up. Wiki files are mirrored into the docs folder, so you can make changes there.

n8detar:
> For now I'd suggest to revert the last commit so the media streaming is working. Then we can look into this further for a future release.

No need to revert, the video still works the same as before.

> You need the following headers: Content-Type,Accept-Encoding, and Range. Note that the last two headers, Accept-Encoding and Range, are additional headers that you may not have needed previously.
> Wildcards "*" cannot be used for the Access-Control-Allow-Origin header. If the page has protected media content, it must use a domain instead of a wildcard.

Is this not something that can be done in the TA nginx config?

bbilly1:
> Is this not something that can be done in the TA nginx config?

That won't be enough. This will have to be passed in dynamically as an environment variable through docker-compose, basically as a host name from where you plan to access the interface. Quite complicated and will require a lot of testing as that can easily break things...

PR #172, opened 2022-02-12
Docs for the cast integration

n8detar:
Includes warnings and requirements for some of the features.


bbilly1:
Hmm, do you think you can attach it here, with the other integrations? 
https://github.com/bbilly1/tubearchivist/blob/master/docs/Settings.md#integrations

So things you can configure on the settings page are mentioned on the settings wiki page too.

Then let's not confuse people with headers and stuff, maybe something like "subtitle support is pending" or coming soon or something like that.

PR #178, opened 2022-02-17
updated redis rejson to use arm64 fork created by bbilly1

schizovivek:


bbilly1:
Sorry, I think there was a misunderstanding. `redislabs/rejson` is the main image for redis and for all amd64 installations, which is probably 90%+. `bbilly1/rejson` is only for arm64, that will not work for other architectures. 
So I think useful would be to have an inline comment there stating that, but keep the redislabs image as the main one and add the bbill1 as the comment. I hope that makes sense?

schizovivek:
yup.. that's what I was trying to ask.. 👍 
I'll update it

bbilly1:
Very nice! Thanks for looking into this.

PR #179, opened 2022-02-18
Save Video Progress

n8detar:
Saves video progress every 10s and on pause.
Progress resets to 0 when the video is marked as watched or unwatched.


bbilly1:
So I have been thinking a bit about the video page problem. On one side it would be quite easy to add that directly to the view and make the variable available in the template.

But I think it would be a good idea to make a API call here too similar as we do with the JS player to get the meta data. This will also help with feature extensions like chapters and stuff like that. And will also reduce code duplication if we can reuse the existing createPlayer function, as far as I can tell, the only change to that function would be, to instead of passing in the button, to just pass the videoId directly.

Do you think that makes sense?

n8detar:
Using the JS would work. Breaking up the create player function wouldn't be a problem. I can begin work on that.

PR #188, opened 2022-03-07
Continue Watching Section

n8detar:
All progress bars are updated on player close
All watch indicators are updated on status change
Added `createWatchStatusIndicator()` function
Replaced `isWatched()` and `isUnwatched()` with `updateVideoWatchStatus()` function

### DIFF
--- a/tubearchivist/home/templates/home/channel_id.html
+++ b/tubearchivist/home/templates/home/channel_id.html
@@ -124,9 +124,9 @@
                     <div class="video-desc {{ view_style }}">
                         <div class="video-desc-player" id="video-info-{{ video.source.youtube_id }}">
                             {% if video.source.player.watched %}
-                                <img src="{% static 'img/icon-seen.svg' %}" alt="seen-icon" data-id="{{ video.source.youtube_id }}" data-status="watched" onclick="isUnwatched(this.id)" class="watch-button" title="Mark as unwatched">
+                                <img src="{% static 'img/icon-seen.svg' %}" alt="seen-icon" data-id="{{ video.source.youtube_id }}" data-status="watched" onclick="updateVideoWatchStatus(this.data-id, this.data-status)" class="watch-button" title="Mark as unwatched">
                             {% else %}
-                                <img src="{% static 'img/icon-unseen.svg' %}" alt="unseen-icon" data-id="{{ video.source.youtube_id }}" data-status="unwatched" onclick="isWatched(this.id)" class="watch-button" title="Mark as watched">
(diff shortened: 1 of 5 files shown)


bbilly1:
Nice, I see you have realized that I missed the video.html template there. :-)

One comment, I think it's more readable to unpack `this` in JS instead of in the template, so for example here:
```
onclick="updateVideoWatchStatus(this.data-id, this.data-status)"
```
you pass multiple arguments into the function. If you do:
```
onclick="updateVideoWatchStatus(this)"
```
and then access the attributes in JS from there, this will simplify the HTML and will also easily make it extendable if need be if we need to pass additional attributes later.

n8detar:
> Nice, I see you have realized that I missed the video.html template there. :-)
> 
> One comment, I think it's more readable to unpack `this` in JS instead of in the template, so for example here:
> 
> ```
> onclick="updateVideoWatchStatus(this.data-id, this.data-status)"
> ```
> 
> you pass multiple arguments into the function. If you do:
> 
> ```
> onclick="updateVideoWatchStatus(this)"
> ```
> 
> and then access the attributes in JS from there, this will simplify the HTML and will also easily make it extendable if need be if we need to pass additional attributes later.

I did it like that so the function could be used in other places, rather than just with the button. If it uses `this` then it needs to be passed the watch button. Maybe we should just use the template to fill in the function like
`onclick="updateVideoWatchStatus({{ video.youtube_id }}, 'unwatched')"`
`onclick="updateVideoWatchStatus({{ video.youtube_id }}, 'watched')"`

n8detar:
> > Nice, I see you have realized that I missed the video.html template there. :-)
> > One comment, I think it's more readable to unpack `this` in JS instead of in the template, so for example here:
> > ```
> > onclick="updateVideoWatchStatus(this.data-id, this.data-status)"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > you pass multiple arguments into the function. If you do:
> > ```
> > onclick="updateVideoWatchStatus(this)"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then access the attributes in JS from there, this will simplify the HTML and will also easily make it extendable if need be if we need to pass additional attributes later.
> 
> I did it like that so the function could be used in other places, rather than just with the button. If it uses `this` then it needs to be passed the watch button. Maybe we should just use the template to fill in the function like `onclick="updateVideoWatchStatus({{ video.youtube_id }}, 'unwatched')"` `onclick="updateVideoWatchStatus({{ video.youtube_id }}, 'watched')"`
> 
> Edit: I thought of another way to do it.

I was planning to pass named arguments, but apparently, JS can't do that so I just added a check to see if the function was passed the button and if there are two inputs to the function.

PR #203, opened 2022-03-29
Modify Dockerfile and the nginx.conf file to avoid Nginx default conf…

Lickitysplitted:
…ig IPv6 conflict.


insuusvenerati:
Why are we wanting to avoid `default.conf`? The `default` config lives in `/etc/nginx/sites-available` 
```
root@b84a9f2ee261:/etc/nginx# tree
.
├── conf.d
├── fastcgi.conf
├── fastcgi_params
├── koi-utf
├── koi-win
├── mime.types
├── modules-available
├── modules-enabled
│   ├── 50-mod-http-geoip.conf -> /usr/share/nginx/modules-available/mod-http-geoip.conf
│   ├── 50-mod-http-image-filter.conf -> /usr/share/nginx/modules-available/mod-http-image-filter.conf
│   ├── 50-mod-http-xslt-filter.conf -> /usr/share/nginx/modules-available/mod-http-xslt-filter.conf
│   ├── 50-mod-mail.conf -> /usr/share/nginx/modules-available/mod-mail.conf
│   ├── 50-mod-stream.conf -> /usr/share/nginx/modules-available/mod-stream.conf
│   └── 70-mod-stream-geoip.conf -> /usr/share/nginx/modules-available/mod-stream-geoip.conf
├── nginx.conf
├── proxy_params
├── scgi_params
├── sites-available
│   └── default
├── sites-enabled
│   └── default -> /etc/nginx/sites-available/default
├── snippets
│   ├── fastcgi-php.conf
│   └── snakeoil.conf
├── uwsgi_params
└── win-utf
```

insuusvenerati:
Ah, just read the rest of the title in the comment. I'm still unsure though, where is the ipv6 conflict?

Lickitysplitted:
In my testing when a host doesn't have IPv6 stack enabled Nginx will fail as the default config listens on IPv6 port 80.

insuusvenerati:
Is it possible to intentionally disable ipv6? I'd like to test this out as well

Lickitysplitted:
Do you mean on the host or inside a container?

bbilly1:
Thanks for getting to work on that so quickly. Two thoughts:

- Do you think it makes sense to keep the filename *nginx.conf* or maybe call it *nginx-default* or so? Just calling it *default* might not be obvious what that file does by just looking at it. You can still copy it to the correct location in the Dockercompose file by specifying the full path.
- I almost forgot, there is a port overwrite with environment variables at startup defined in *run.sh* here: https://github.com/bbilly1/tubearchivist/blob/master/docker_assets/run.sh#L16-L24, that's a workaround for some edge cases where the docker internal networking is shared with the host and [ports collide](https://github.com/bbilly1/tubearchivist#port-collisions), so this will need changing there too.

Thanks a lot for your help! You can just push any further changes to your fork, this PR should update automatically.

Lickitysplitted:
> * Do you think it makes sense to keep the filename _nginx.conf_ or maybe call it _nginx-default_ or so? Just calling it _default_ might not be obvious what that file does by just looking at it.

You make a good point, but in a single step in the Dockerfile we are able to overwrite the config file that causes the problem and place the desired config file in place. Otherwise an additional step would be needed to create a symlink or enable the config into the sites-enabled folder. I know the current image is really heavy and optimizing the Dockerfile can help.

Is there an expectation that users will modify or need to know about this config file? If devs need to know it can be made in the 
 contribution documentation.

Lickitysplitted:
> * I almost forgot, there is a port overwrite with environment variables at startup defined in _run.sh_ here: https://github.com/bbilly1/tubearchivist/blob/master/docker_assets/run.sh#L16-L24, that's a workaround for some edge cases where the docker internal networking is shared with the host and [ports collide](https://github.com/bbilly1/tubearchivist#port-collisions), so this will need changing there too.

Does this fix the issue where nginx is listening on port 80 on both IPv4 and IPv6 by default?

Lickitysplitted:
> I assume on the host since that's what you mentioned in your previous comment.

So generally speaking there are two accepted methods of disabling IPv6 on a Linux host. One is to disable within the grub cmd and this results in no IPv6 sysctl options at all. The other is to use sysctl options which has similar results but the sysctl options still remain. AFAIK docker will adhere to the networking that is available on the host, but nginx inside of a container doesn't know that.
Since nginx is getting installed with defaults it places the default config file in the sites-available with the symlink enabled to the sites-enabled. The behavior of this default config file is to listen on port 80 on both IPv4 and IPv6. If you go exec into your container now you will probably see a listener on port 80.
If a system does not have IPv6 kernel or system options available this nginx config will fail causing nginx to fail to load. This is what I've observed and my host has IPv6 disabled at grub. Reproducing this with VMs shouldn't be difficult.

bbilly1:
Just double checked things, it looks like we lost docker_assets/nginx.conf now. I think there went something wrong when renaming the file and it got accidentally deleted?

bbilly1:
Took the liberty to add the file back. Thanks a lot for your help with that! 

And congratulations to your first merged pull request! :-)

insuusvenerati:
Ooh interesting. Do you know of a way to intentionally disable the ipv6 stack? I’d like to test as well

Issue #210, opened 2022-04-05
Download progress box disappears 


fromnack:
Hi there

Sorry if I've not done this right, I've never submitted a bug report on Github before! 

Running on the latest version on Unraid. 

The download progress box disappears randomly, but downloads are still going. It comes and goes intermittently with nothing specific bringing it back, sometimes a refresh or changing pages brings it back. 

Let me know if you need more info!

ainsey11:
This appears to happen when the file takes longer than the message expiry in redis

I'm experimenting putting in a UI box to show that the file move is ongoing, something a little like this:

![image](https://user-images.githubusercontent.com/6608275/162634552-e5d9d120-a8cf-4631-baec-58bc329686a8.png)

ainsey11:
PR has been updated ready for review.

PR #219, opened 2022-04-10
Fix: #210 - Download UI Box being removed when file is still processing

ainsey11:
The UI element is removed when the file has finished downloading, but is still being moved on disk to the storage destination. This means the user is presented with nothing in the UI yet the file is still being processed.

This PR is to add in a message just before the move_to_archive() function is called, with no expiry, so the webUI updates with a "Moving file" UI element so the user can see that the file is still in flight.

image

Eventually I would like to display the video name of the file being moved, but keeping large changes to a minimum whislt I still find my feet in the project

### DIFF
--- a/tubearchivist/home/src/download/yt_dlp_handler.py
+++ b/tubearchivist/home/src/download/yt_dlp_handler.py
@@ -187,9 +187,16 @@ class VideoDownloader:
                 "title": "Moving....",
                 "message": "Moving downloaded file to storage folder",
             }
-            RedisArchivist().set_message("message:download", mess_dict)
+            RedisArchivist().set_message("message:download", mess_dict, False)
 
             self.move_to_archive(vid_dict)
+            mess_dict = {
+                "status": "message:download",
+                "level": "info",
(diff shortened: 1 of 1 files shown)


ainsey11:
PR to resolve https://github.com/bbilly1/tubearchivist/issues/210

bbilly1:
Nice, I think thats the right approach. 

So for "historical" reason, the set_message expires by default, so you can either pass a number as seconds, or False to make it permanent. I kjow, I know... J need to start write some doc strings on these classes...

So I was thinking, how about at the beginning of the move task, set a message that doesnt expire, and at the end send a message that maybe expires in 5secs or so? Then when the task completes, the message box doesnt linger too long?

ainsey11:
Thanks @bbilly1 

I'll tweak and append a new commit in later today

Cheers
Rob

ainsey11:
Updated to set a message on the move that never expires, followed by a message after the move with a 4 second expiry that states "completed"

Looks like this:
image

updated PR incoming

PR #222, opened 2022-04-12
Adjust to API changes for SB integration.

n8detar:


bbilly1:
Excellent! I see this has simplified things. Thanks for your help!

PR #225, opened 2022-04-14
chore: add base64 image info to video API

insuusvenerati:


bbilly1:
This shouldn't be necessary. The `vid_thumb_base64` field already gets returned automatically in the API. Can it be that you have videos downloaded and indexed from before? Only new videos will have that field added automatically. 

You could delete `ta_video` from ES, restart the container and run `scan filesystem` to index all with the newest mappings. Then before we can ship that, we'll have to implement a startup check to add the base64 image to all videos. But this should then also incorporate channel and playlist art work too.

PR #233, opened 2022-04-21
feat(auth): add REMOTE_USER authentication

iarp:
Enable django's REMOTE_USER middleware and authentication backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ The main Python application that displays and serves your video collection, buil
   - Change the environment variables `TA_USERNAME` and `TA_PASSWORD` to create the initial credentials. 
   - `ELASTIC_PASSWORD` is for the password for Elasticsearch. The environment variable `ELASTIC_USER` is optional, should you want to change the username from the default *elastic*.
   - For the scheduler to know what time it is, set your timezone with the `TZ` environment variable, defaults to *UTC*.
+  - `DJANGO_REMOTE_USER_HEADER="REMOTE_USER"` changes the header used in django RemoteUserBackend, `DJANGO_REMOTE_USER_CREATE_UNKNOWN_USER=True or False` allows automatic user creation if the matching remote_user username does not match a user in django.
 
 ### Port collisions
 If you have a collision on port `8000`, best solution is to use dockers *HOST_PORT* and *CONTAINER_PORT* distinction: To for example change the interface to port 9000 use `9000:8000` in your docker-compose file.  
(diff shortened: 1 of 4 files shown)


bbilly1:
Nice! Thanks for working on that. 

The linter is complaining, you can run `deploy.sh validate` to validate your changes: https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#code-formatting-and-linting, just so we all use the same formatting and style to try to keep things consistent. 

Then this will also need documentation what it does and on how to use that feature. Not sure where that would fit best, maybe here: https://github.com/tubearchivist/tubearchivist#tube-archivist? Or what do you think?

bbilly1:
That wiki page is still WIP, for unraid only for now. I think it will fit best with the other environment variables in the Readme. 

Maybe make a separate PR for that? So I can merge that before making a new release, so people don't get confused as this won't be in their installation just yet.

bbilly1:
I think I lost you on Discord. Do you mind helping me understand how this works in detail? It's honestly my first time messing around with custom auth classes and I could use your help with making sure I understand it properly.

iarp:
When you use a reverse proxy in front of TA you can setup a single sign-on service to act as authentication for all your apps in one location rather than having to login to an account on each app.

In my use case I use nginx as my reverse proxy and [Authelia](https://github.com/authelia/authelia) (SSO) as the authentication. When a request comes in nginx contacts the SSO server to confirm my browser session is logged into SSO. SSO returns the username, email, and groups the SSO user is in. Nginx takes these values and puts them into the headers REMOTE_USER, REMOTE_EMAIL, REMOTE_GROUPS.

You can then use these values to login/register a user as they visit the page as long as they're authenticated with the SSO server.

In the case of this PR, django's authentication is setup to check for REMOTE_USER first and falls back to the usual authentication. This allows someone that is not authenticated with SSO to still login using regular django authentication.

I will leave the decision up to the rest of you as to whether or not you want to merge or just close this. At this time I've decided to look into other potential media options as the overhead on my server was becoming too much, I left the discord after I had started seeing conversations of adding even more containers.

bbilly1:
I assume you are referring to our discussion on how to best package the new frontend? Lot's of pros and cons to consider, that's what Discord if for. As these things go, sometimes ideas are getting discussed that don't end up being realized as somebody comes up with a better solution, that's the beauty of collaboration. Looks like we could use your voice and input!

So for this PR: I'm happy to merge it but I need to know if you are on board with this project. I'm sure that's not the last time we need to tinker with authentication and I don't want to end up in a position of having a feature that I'm not using and I'm not able to support. That's not saying I don't see the benefit of SSO, thank you for your explanation.

bbilly1:
Closing this for now. If somebody wants to maintain that in the future, please reach out and we'll reopen. Thank you for your understanding.

PR #236, opened 2022-04-30
Fix to prevent holding onto SB info.

n8detar:
I couldn't replicate the issue, but this should prevent it from happening.


bbilly1:
This does indeed fix it. Thanks for looking into it! :-)

PR #242, opened 2022-05-03
Cookie Validation Button

n8detar:


bbilly1:
Excellent, thanks for working on that. I added a little if clause to only show the validate button if cookie is enabled. and a *processing* message to get immediate feedback when you click as it can take a sec or two until you get the response back.

PR #251, opened 2022-05-21
Add Truenas Scale Instructions

Heavybullets8:
included two 
```
<br />
```
 to add spacing between the guides. 
If the website results in an ssl error at the time of submitting its because the new site is still in the process of being handed over to a different registrar, and its still getting its certs from github. Should be fixed by tomorrow.


bbilly1:
Very nice, thanks for adding this here. Let's give this a day to propagate and I'll merge it tomorrow.

Issue #257, opened 2022-06-06
API: TaskAPIView does not include a GET to request active tasks


lamusmaser:
TaskAPIView only includes the POST request to run a new task, but does not have a method to request if there is an ongoing task, and the POST response adds to the the requested task to the queue of tasks, if able.

Request:
Add a GET call to TaskAPIView that will return either the active task (optional: current stage ongoing task is at) or a POST call that requests if that type of task is currently ongoing.

lamusmaser:
I can look into making this into a PR if I get some time, but wanted to make sure the request was made here in case someone else is able to get to it before I can.

bbilly1:
So far, you can see your running tasks in `/progress/`, that's also how the notifications are built. But yes, for a rest full approach, this should be done in `/api/task/`. Then a GET request would return a list of messages? Associated with a task or a interaction from the frontend? I wanted to avoid making dedicated endpoints for each task, e.g. `/api/task/download_pending/` etc to not have this extend endlessly, but not sure if that is the right approach. 

So I'd be very happy if you could look into this.

lamusmaser:
I think having the one task response (since only one task can be running at a time) should be fine. The current method of making a request to `/progress` is not as compatible with the current REST API requests, since you have to login via the `/login/` and pickup the CSRF token from there to get it to work properly.

lamusmaser:
Will be closed with next version; PR accepted and merged successfully.

PR #258, opened 2022-06-06
Update TaskAPIView with GET - #257 PR

lamusmaser:
Initial commit to determine the GET response for TaskAPIView which should show if the "rescan" or "download" tasks are currently running. This is based on if the Lock is currently owned - may need to modify further if determined inadequate for task checking.

### DIFF
--- a/tubearchivist/api/README.md
+++ b/tubearchivist/api/README.md
@@ -211,8 +211,8 @@ GET /api/task/
 Returns:
 ```json
 {
-    "rescan": "false",
-    "download": "false"
+    "rescan": false,
+    "download": false
 }
 ```
 
--- a/tubearchivist/api/README.md
+++ b/tubearchivist/api/README.md
@@ -202,6 +202,20 @@ When valid returns message with user id:
 ```
 
 ## Task View
+GET /api/task/
+POST /api/task/
+
+Check if there is an ongoing task:
+GET /api/task/
+
+Returns:
+```json
+{
+    "rescan": "false",
+    "download": "false"
+}
+```
+
 Start a background task
 POST /api/task/
 ```json
--- a/tubearchivist/api/README.md
+++ b/tubearchivist/api/README.md
@@ -212,7 +212,7 @@ Returns:
 ```json
 {
     "rescan": false,
-    "download": false
+    "downloading": false
 }
 ```
 
--- a/tubearchivist/api/views.py
+++ b/tubearchivist/api/views.py
@@ -456,6 +456,10 @@ class TaskApiView(ApiBaseView):
     POST: start a new background task
     """
 
+    @staticmethod
+    def get(request):
+        return Response({"Test": "success"})
+
     # @staticmethod
     # def get(request):
     #     """handle get request"""
--- a/tubearchivist/api/views.py
+++ b/tubearchivist/api/views.py
@@ -458,7 +458,7 @@ class TaskApiView(ApiBaseView):
 
     def get(request):
         """handle get request"""
-        response = {"rescan": False, "download":False}
+        response = {"rescan": False, "download": False}
         response["rescan"] = RedisArchivist().get_lock("rescan").owned()
         response["download"] = RedisArchivist().get_lock("rescan").owned()
 
--- a/tubearchivist/api/views.py
+++ b/tubearchivist/api/views.py
@@ -455,7 +455,7 @@ class TaskApiView(ApiBaseView):
     GET: check if ongoing background task
     POST: start a new background task
     """
-    
+
     @staticmethod
     def get(request):
         """handle get request"""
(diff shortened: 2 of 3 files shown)


lamusmaser:
This is associated with #257 and will close when merged.

bbilly1:
Thanks for getting to work on that. The method `get_lock` of the `RedisArchivist` class will return a new lock instance every time it's called, so you won't be able to run `owned()` on a different instance than where the lock was initiated. I hope that makes sense, basically you can only call the `owned()` method here: https://github.com/tubearchivist/tubearchivist/blob/bb0f3dad1a02a375a306cc65b1689de92b6fc8e8/tubearchivist/home/tasks.py#L54 and it will be doing what you think it does. Otherwise you'll have to `acquire()` it, but that will create it if it can, so that's not what you want either. 

Basically a lock is just a key set in redis, so you can check if that key is there, that key will automatically disappear when `release()` was called on that lock, so that is a thread independent approach. 

I've pushed a commit to your PR implementing `is_locked` in the `RedisArchivist` class that is just doing that, so you can call that method in the API view.

Also note that all API views expect `self` as it's first argument and `request` as the second, like here: https://github.com/tubearchivist/tubearchivist/blob/bb0f3dad1a02a375a306cc65b1689de92b6fc8e8/tubearchivist/api/views.py#L366 you'll have to use the `staticmethod` decorator if you don't need access to `self` in that view, like here: https://github.com/tubearchivist/tubearchivist/blob/bb0f3dad1a02a375a306cc65b1689de92b6fc8e8/tubearchivist/api/views.py#L427

Otherwise Django will be unhappy because there is an unexpected amount of arguments.

lamusmaser:
Thank you for the detailed reply. Based on my initial reading, I thought that `.get_lock()` was doing what the new `is_locked()` function is doing, so I appreciate the commit. I'll update the PR with the replacement requests.

Also noted for the `self` vs `staticmethod` decorator; I had noticed what you pointed out, but didn't make the connection that the decorator was what allowed it.

bbilly1:
Were you able to build the image and test your changes? For linting, you can use `./deploy.sh validate` to help you out there.

lamusmaser:
I haven't yet. I had to shut down my test server a while back, so I didn't get the time to redeploy and rebuild the image.

I will see if I can get that done sometime later this week before the PR merge.

lamusmaser:
@bbilly1 - I have this tested and working now. Looks like we were getting a key error, but now it responds appropriately:

```
>>> r = requests.post(base_url+login_api,json=payload)
>>> headers = {"Authorization": f"Token {r.json()['token']}"}
>>> p = requests.get(base_url+ping_url,headers=headers)
>>> p.text
'{"response":"pong","user":1}'
>>> tg = requests.get(base_url+task_api,headers=headers)
>>> tg.text
'{"rescan":false,"download":false}'
>>> tp = requests.post(base_url+task_api,headers=headers,json=runtask)
>>> tp.text
'{"success":true,"task":"rescan_pending"}'
```

Testing also shows that the `POST` requests are no longer failing with Server 500 responses.

bbilly1:
Thanks for working on that. Does it ever set to `true` for you when you have the rescan process running for example? You'll have to pass the name of the lock into the `is_locked` method. Like here: https://github.com/tubearchivist/tubearchivist/blob/bb0f3dad1a02a375a306cc65b1689de92b6fc8e8/tubearchivist/home/tasks.py#L82 this creates a lock named `downloading` so to check if that lock is set, you'll have to use the same name like `RedisArchivist().is_locked("downloading")`.

Then I saw, you also updated the README, good thinking. The return will be a bool though and not a string, so `"download": false` and not `"download": "false"`.

lamusmaser:
README has been updated appropriately.

Had to make a minor adjustment - I was making the correct call, but wasn't using the correct key (misread which string was passed through to the end). 

```
>>> tg = requests.get(base_url+task_url,headers=headers)
>>> tg.text
'{"rescan":false,"download":false}'
>>> tp = requests.post(base_url+task_url,headers=headers,json=runtask)
>>> tp.text
'{"success":true,"task":"rescan_pending"}'
>>> tg = requests.get(base_url+task_url,headers=headers)
>>> tg.text
'{"rescan":true,"download":false}'
>>> >>> tg = requests.get(base_url+task_url,headers=headers)
>>> tg.text
'{"rescan":false,"download":false}'
```

Download was not working, so pushed another update that 1) fixed the `download` key to be `downloading` and 2) converted it to a dictionary reference loop, rather than keying each individually.

```
>>> tp = requests.post(base_url+task_url,headers=headers,json=runtask)
>>> tp.text
'{"success":true,"task":"download_pending"}'
>>> tg = requests.get(base_url+task_url,headers=headers)
>>> tg.text
'{"rescan":false,"downloading":true}'
```

At this point, the endpoint is working and responding against the two tasks; additionally, the README should be correct with the right typecast, as well.

bbilly1:
Excellent work, this can then easily be extended with new tasks by just adding them to the list.

PR #263, opened 2022-06-18
Fix repeated superuser creations after rename/removal

### LINKED ISSUE #261: [docker] initial superuser created every time container starts
So it appears that the superuser is created on container startup every time based on variables `TA_USERNAME` and `TA_PASSWORD`, even if:
- the container is started for the non-first time
- even if another superuser exists
- even if the initial user is manually deleted using the admin interface

One would presume from the wording on README, i.e. `Change the environment variables TA_USERNAME and TA_PASSWORD to create the initial credentials` and the example compose file, i.e. `your initial TA credentials` that it's only created when the container is run for the first time and/or if no other superuser exists. I believe this should be more clear in the README, and/or make the app not create a superuser if it's not the first time being run

dshoreman:
This prevents repeated runs of `createsuperuser` after the first successful one.  

* It uses a file to determine whether creation has already been run.
* Currently that file is `/cache/initsu.lock`, or `./initsu.lock` when outside Docker.
* If the file exists, it skips creation with a warning (and instructions to force it if that's what the user wants).
* If it *doesn't*, it's assumed to be a new install; creation happens as normal, file created on success.

---
#### Some notes
1. `ELASTIC_USER` is no longer required:
   * The original requirement had no effect anyway because of the 'elastic' default
2. Other variables will now error if they're set-but-empty:
   * They could previously be *empty*, as long as they were *set*
   * As an added bonus, this also fixes a fun accidental lockout scenario: you could *create* an account with no password, but it won't let you *log in* with one due to input validation on the login page.

**What this PR *does not* do:**
* *Account for existing installs where the default was removed/renamed*
  One final creation is inevitable, but could optionally be avoided with `touch cache/initsu.lock` before starting.
* *Enable any automatic creation after the first successful run* (...but that may be a feature in itself)  
  Shouldn't be needed, but steps to force re-creation are printed out to the log if someone *really* wants to

Fixes #261

### DIFF
--- a/docker_assets/run.sh
+++ b/docker_assets/run.sh
@@ -5,7 +5,12 @@ if [[ -z "$ELASTIC_USER" ]]; then
     export ELASTIC_USER=elastic
 fi
 
-: "${ELASTIC_PASSWORD:?"Missing required environment variable"}"
+required="Missing required environment variable"
+if [[ ! -f .superuser_created ]]; then
+    : "${TA_USERNAME:?$required}"
+    : "${TA_PASSWORD?$required}"
+fi
+: "${ELASTIC_PASSWORD:?$required}"
 
 # ugly nginx and uwsgi port overwrite with env vars
 if [[ -n "$TA_PORT" ]]; then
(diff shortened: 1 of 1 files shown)


dshoreman:
I ended up having to parse output of `createsuperuser`. If the user already exists (eg in first post-upgrade starts), the command throws an error into stderr and exits with a non-zero status so the temp file would never be created.

Seems to be working for me now, description updated! :tada:

Issue #268, opened 2022-07-14
Videos won't play (403 Forbidden) due to how nginx is configured


p0358:
This is potentially the same issue as in #5, except I found the exact culprit.

The issue is that everything in the container runs as `root` (fine on its own), but nginx runs as `www-data`. So I ended up with an counter-intuitive situation, where the app would successfully download videos and its metadata, but attempting to play them back would result in a broken player in every browser. Opening webtools I've quickly learnt that the culprit is not the lack of codec, as Firefox reported in the UI, but error 403 from nginx. I've installed procps on the container and seen that nginx runs as www-data:

```
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   5728  1704 ?        Ss   12:23   0:00 /bin/bash ./run.sh
root          16  0.2  0.4 122480 65772 ?        S    12:23   0:03 /usr/local/bin/python /root/.local/bin/celery -A home.tasks worker --loglevel=INFO
root          17  0.0  0.3 134616 65000 ?        S    12:23   0:00 /usr/local/bin/python /root/.local/bin/celery -A home beat --loglevel=INFO -s /cache/celerybeat-schedule
root          18  0.0  0.3 112976 54556 ?        S    12:23   0:00 uwsgi --ini uwsgi.ini
root          19  0.0  0.0  58352  1568 ?        Ss   12:23   0:00 nginx: master process nginx
www-data      20  0.0  0.0  59036  3788 ?        S    12:23   0:00 nginx: worker process
www-data      21  0.0  0.0  58684  2820 ?        S    12:23   0:00 nginx: worker process
www-data      22  0.0  0.0  58684  2820 ?        S    12:23   0:00 nginx: worker process
www-data      23  0.0  0.0  58684  2820 ?        S    12:23   0:00 nginx: worker process
root          27  0.1  0.3 121928 62712 ?        S    12:23   0:02 uwsgi --ini uwsgi.ini
root          29  0.0  0.3 122196 61216 ?        S    12:23   0:00 /usr/local/bin/python /root/.local/bin/celery -A home.tasks worker --loglevel=INFO
root          30  0.0  0.3 122200 61260 ?        S    12:23   0:00 /usr/local/bin/python /root/.local/bin/celery -A home.tasks worker --loglevel=INFO
[...]

PR #270, opened 2022-07-14
Fix permission/mount issues from #268

p0358:
1. nginx is ran as root, the same as the main app and its services. This alleviates the issue in which nginx was unable to read the /youtube folder due to running as arbitrary user www-data unknown to the host. The Docker mounts, when using a host path, replicate directly the host permissions and owners UID and GID, which is why it wasn't working, resulting in an odd behavior from the user perspective, because downloading videos itself worked just fine.
I also took the liberty to change three `RUN` instructions that create dirs into just one for performance reasons, as every separate `RUN` directive causes a new layer to be created.

2. Changed
```py
shutil.move(old_file_path, new_file_path)
```
to
```
shutil.move(old_file_path, new_file_path, copy_function=shutil.copy)
```
as described on the issue page, as by default `shutil.copy2` fallback copy function is used, which tries to copy all the metadata and permissions from the source (/cache), which in this case is undesired and may (and does) lead to problems when the mounts are on different filesystems. Instead `shutil.copy` uses standard `cp` semantics, without trying to replicate the metadata after the copy.

I did keep the `os.chown` call below, as that seems to work just fine on the other hand, and is actually a useful functionality.

There are two more `shutil.move` calls across the whole codebase, but it is my understanding (might be wrong) that they operate only within the /cache dirs, and as such might be fine with no changes to them.

I tested these two changes on docker-compose setup on Debian and they seem to cause no side effects. You can reproduce the nginx problem by having a volume like `./youtube:/youtube` with permissions `700` on `./youtube`, owned by `root:root`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,12 +54,11 @@ RUN if [ "$INSTALL_DEBUG" ] ; then \
     ; fi
 
 # make folders
-RUN mkdir /cache
-RUN mkdir /youtube
-RUN mkdir /app
+RUN mkdir /cache /youtube /app
 
 # copy config files
 COPY docker_assets/nginx.conf /etc/nginx/sites-available/default
+RUN sed -i 's/^user www\-data\;$/user root\;/' /etc/nginx/nginx.conf
 
 # copy application into container
 COPY ./tubearchivist /app
--- a/tubearchivist/home/src/download/yt_dlp_handler.py
+++ b/tubearchivist/home/src/download/yt_dlp_handler.py
@@ -398,7 +398,7 @@ class VideoDownloader:
         old_file_path = os.path.join(cache_dir, "download", old_file)
         new_file_path = os.path.join(videos, vid_dict["media_url"])
         # move media file and fix permission
-        shutil.move(old_file_path, new_file_path)
+        shutil.move(old_file_path, new_file_path, copy_function=shutil.copy)
         if host_uid and host_gid:
             os.chown(new_file_path, host_uid, host_gid)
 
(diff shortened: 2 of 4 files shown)


bbilly1:
Thank you for preparing this.

> but it is my understanding (might be wrong) that they operate only within the /cache dirs  

That's only partially correct: One for manual media ingest, to move the media file from `/cache/import` to `/cache/download`, to then use the `VideoDownloader.move_to_archive` method to reuse the rest of the class: https://github.com/tubearchivist/tubearchivist/blob/1479516689badaea804ab9bda52b4f27bdf5109b/tubearchivist/home/src/index/filesystem.py#L269

But the other one is to catch renamed channels in the `Reindex` class. That's when the uploader changes the name, this will have to be reflected in TA as well in the channel folder name. So this `ChannelUrlFixer` class in the reindex module catches these cases, if it finds any issue, moves the video file out from `/youtube` to `/cache/download` to then use the rest of the `move_to_archive` function to put it back into the right place. I know, maybe not very elegant... But makes all the code reusable. :-) So that would be between mount points and for that, would have to use `copy_function=shutil.copy` too: https://github.com/tubearchivist/tubearchivist/blob/1479516689badaea804ab9bda52b4f27bdf5109b/tubearchivist/home/src/index/reindex.py#L307

Thank you for your hard work in that!

And excellent catch on the `RUN mkdir` call in the Dockerfile, I wasn't aware that you can join them together like that. 

For nginx, that didn't have any issues on my testing, so I think that solution works.

bbilly1:
For the sake of completeness, from Discord: adding `copy_function=shutil.copy` to all `shutil.move` calls for consistency and to catch all cases.

Issue #78, opened 2021-11-01
Add sso / ldap support


mindofbeholder:
My use case is that I utilize Authentik for SSO / LDAP.

I want to be able to automatically create my users in tubearchivist when they authenticate from my authentik instance.

bbilly1:
Yes, something to look into further down the road once the user and user roles situation is clear.

bbilly1:
I've added it to the roadmap for now but honestly I'm not sure if and when I'll be able to look at that, as this is something I'm not using my self. If you want to get this going, please contribute!

Closing this issue for now.

PR #274, opened 2022-07-23
Add LDAP authentication support

DanielBatteryStapler:
Hello! I really like this project but there was just one feature missing that I would like, so I added to my hosted version. It seems to work just fine, so here's the pull request with the changes. All I did was use some environment variables that can be specified in the docker-compose.yml file to pass the correct settings in setting.py to Django to enable and configure LDAP. I had to add some more libraries in the Dockerfile and requirements.txt files to be installed in order for Django to have LDAP support enabled and working.

There currently isn't any configuration options for user roles because user roles haven't landed in tubearchivist, but once they do I'll add support for those through LDAP too.

I've added some documentation to the FAQ file specifying what the different environment variables do with some example values in place of the section saying that there isn't LDAP support.

This pull request will also resolve issue #78.

Let me know if there are any changes I need to make to make this merge-worthy, and thank you again for making tubearchivist!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ FROM python:3.10.5-slim-bullseye AS builder
 ARG TARGETPLATFORM
 
 RUN apt-get update
-RUN apt-get install -y --no-install-recommends build-essential gcc
+RUN apt-get install -y --no-install-recommends build-essential gcc libldap2-dev libsasl2-dev libssl-dev
 
 # install requirements
 COPY ./tubearchivist/requirements.txt /requirements.txt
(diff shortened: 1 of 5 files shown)


bbilly1:
Hi there, thanks a lot for working on that! I was about to wrap this up into a new release, as there are quite a lot of improvements waiting for publishing. 

I'll look into this as soon as I can. Also thank you for reaching out on Discord! :-)

bbilly1:
Just to add the summary from our Discord conversation. Thanks a lot for working on that. What we have discussed: 
- Documentation of the env variables would fit best under the installation instructions here: https://github.com/tubearchivist/tubearchivist#installing-and-updating
- Add a mention that this is in testing and not yet available in regular release.

Then this can be merged and tested with an *unstable* build.

Issue #290, opened 2022-07-29
Missing search API endpoint


PrivateGER:
While trying to write my own client for Tubearchivist, I noticed that there isn't any sort of API endpoint for search.

The /process/ endpoint does the job perfectly, but unfortunately requires a CSRF token and therefore does not work with token-based auth.

Can we get a search API endpoint, or allow token-auth for /process/?

bbilly1:
Yeah, the API is under development and far from complete and [this here](https://github.com/tubearchivist/tubearchivist/tree/master/tubearchivist/api) is very much true: 
> This is very early stages and will change!  

At current state, there are quite a few things missing or only partially implemented, search being only one of them.

Did you see, there is already a good amount of progress with the NextJS/React frontend: https://github.com/tubearchivist/tubearchivist-frontend, but we stalled there, due to life but is just begging you to pickup where we left off! :-)

The `/process/` url was a *poor mans API* in the beginning, I have been moving things over to to the *real* api endpoints. 

In any case, please reach out on Discord, I'd love to hear your plans! And please help with developing the missing pieces there.

bbilly1:
Thanks for implementing that. PR is now out in v0.2.1.

PR #303, opened 2022-08-10
Add API endpoint for searching 

PrivateGER:
Created for #290  

This adds the /api/search?query=term endpoint to facilitate API-based search. I also changed the frontend to use this new endpoint.

### DIFF
--- a/tubearchivist/api/urls.py
+++ b/tubearchivist/api/urls.py
@@ -1,7 +1,5 @@
 """all api urls"""
 
-from django.urls import path
-
 from api.views import (
     ChannelApiListView,
     ChannelApiVideoView,
@@ -21,6 +19,7 @@ from api.views import (
     VideoProgressView,
     VideoSponsorView,
 )
+from django.urls import path
 
 urlpatterns = [
     path("ping/", PingView.as_view(), name="ping"),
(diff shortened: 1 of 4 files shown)


PrivateGER:
The CI linter thingy seems to not like my formatting, even though it passes all linters just fine locally. No idea what that's about.

bbilly1:
Excellent work, thank you for working on that! 

To make the linter happy, you can use the `deploy.sh validate` script, that runs all the same linter as the action does, so you get immediate and exact feedback: https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#code-formatting-and-linting. 

You did the work so you should get the honors: You can remove the old method from the *api_calls.py* module:
https://github.com/tubearchivist/tubearchivist/blob/13018192f327f18e17ac1cf7c1d05afc6a6c76f9/tubearchivist/home/src/frontend/api_calls.py#L289-L294

One more done, only 20 left to go. :-)

I've added a few more comments in the code.

bbilly1:
Were you able to look into this? So we can wrap this up before the next release.

Otherwise, if you prefere, I can also push the changes to your PR.

PrivateGER:
Was a bit busy recently, this should solve your comments.

PR #340, opened 2022-10-22
More detailed install instructions

bakkot:
The install instructions assume you already know what to do with a docker-compose.yml file, and aren't super clear about what things from the docker-compose file you'll need to edit. This revision hopefully makes it a little easier to follow if you haven't run into docker before.

Please confirm the list of what parts of the docker-compose file need to be edited; I'm not totally sure I've got everything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![Tube Archivist](assets/tube-archivist-banner.jpg?raw=true "Tube Archivist Banner")
+![Tube Archivist](assets/tube-archivist-banner.jpg?raw=true "Tube Archivist Banner")  
 
 <h1 align="center">Your self hosted YouTube media server</h1>
 <div align="center">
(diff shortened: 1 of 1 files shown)


lamusmaser:
Definitely a good idea, however I think most of this information should be focused in the wiki, rather than the README. Primary reason for thinking this: while Docker installations are primarily focused on *UNIX-based platforms, there is a large enough portion of those interested in trying out the project that come from Windows-based platforms or are not as familiar with the requirements for setting up the environment that would particular to their use-case and implementation.

I do think this should be appropriately documented, I think we should have some high-level responses for certain platforms. I do know there was some discussion on the Discord server regarding creating documentation for Docker/Docker-compose on Linux systems that were not supported by an underlying utility (as we already have documented in the wiki). Statement from @bbilly1 was that we shouldn't reinvent the wheel unless we were willing to maintain the portions of documentation related to installation _if those instructions should change_ or making larger assumptions about environments (do we need to create RPM vs DEB based installation instructions; do we need to create Arch Linux instructions; for the RHEL/Fedora family, do we assume that they have access to the packages or use the install directions provided by Docker; etc.). 

The assumptions that are made with this are that one of the latest Docker versions is both available and installed (as seen from using the `docker compose` command strings, which are a fairly recent addition). I don't disagree with creating the assumption, but we do need to ask if we should accommodate for older version iterations, what is our version compatibility we are maintaining (and our deprecation process for older versions), and how we appropriately document these details, regardless if they are in the README vs wiki.

@bbilly1, you ultimately approve on these and provide responses, but those are just the primary things I wanted to comment on based on what we had previously discussed regarding generating documentation for installation.

bakkot:
I think the readme should tell you how to do the install. It doesn't currently - it says "look at the example `docker-compose.yml` provided", which is not really useful if you aren't already familiar with docker. In particular I think the readme definitely should cover

- you need to have docker installed (not _how_ to install it, which the rest of the internet can help with once you know that's what you need)
- you _only_ need the docker-compose file, not the rest of the repository (which is not usually the case even for docker-based projects, in my experience)
- you will need to edit certain field in the docker-compose file, and
- which fields you need to edit

That's pretty much all this PR does.

---

Regarding `docker compose` assuming a recent version of docker - I think anyone who has an old version of docker installed already will probably not need the readme to tell them how to user a docker-compose file, so it's not going to be confusing to them. The commands are just for people new to docker, who presumably will be installing the latest version, which supports `docker compose`. I don't think that having the install instructions list the current version of the command really requires establishing a policy for compatibility with old versions.

bbilly1:
First, thank you from getting the ball rolling here, I really appreciate it.

How about we have on the readme a short and concise installation instructions, where folks who have a reasonable understanding of docker and docker compose can get a quick overview from the various configuration options? So they should be able to get up and running quickly with the example docker-compose.yml file without having to go through a long set of instructions more geared towards beginner?

Then we'll have a more *generic*, *extensive* and *beginner friendly* set of instructions in the [wiki](https://github.com/tubearchivist/tubearchivist/wiki/Installation)? Basically beside the existing platforms, we can also have a *linux* section. This way we can keep the readme as the frontpage short and concise, but if people need more help, we can have that all on the wiki page? This would also keep us from duplicating - at least partially - the existing instructions, [here](https://github.com/tubearchivist/tubearchivist#tube-archivist) for example. 

Additionally, there are also some helpful comments in the [docker compose file](https://github.com/tubearchivist/tubearchivist/blob/master/docker-compose.yml), mostly from errors and wrong configurations we have seen.

What do you think? But all in all, I hope this can be a community effort, where all contribute to the documentation improving things and make sure it's complete and up to date.

bakkot:
That all sounds good, but "short and concise install instructions [with] a quick overview of the various configuration options" was what I was going for here. If you want something different than what's in this PR I'm not sure what changes you'd like to see.

I definitely agree that stuff like "how to set up docker" shouldn't be in scope for the readme, but the basic "you need to set up docker" and "you need to save the docker-compose.yml file from this repo, rather than cloning the whole repo" and "you will need to change these parts of the sample config" stuff seems like it's the bare minimum that should be covered. It doesn't need to tell you _how_ to do that stuff, but it should at least tell you _that_ you need to do that stuff.

(That said, I don't think it's a good idea to assume the reader is already familiar with docker. Most people aren't. And even for people who are, many docker-based projects require setup beyond just `docker compose up`, so it's still important to point out that this is all you need to do.)

bbilly1:
Excellent, thanks for working on that!

Issue #341, opened 2022-10-22
[Feature Request]: keyboard shortcuts in the player


bakkot:
### Already implemented?

- [X] I have read through the [wiki](https://github.com/tubearchivist/tubearchivist/wiki).
- [X] I understand the [scope](https://github.com/tubearchivist/tubearchivist/wiki/FAQ) of this project and am aware of the [known limitations](https://github.com/tubearchivist/tubearchivist#known-limitations) and my idea is not already on the [roadmap](https://github.com/tubearchivist/tubearchivist#roadmap).

### Your Feature Request

I'd like to have at least some keyboard shortcuts available on the video player. The ones I use on Youtube all the time are:

- `>`/`<` to change video speed (maybe going up to 3x instead of 2x?)
- `c` to toggle captions
- left and right arrow to skip around in 5 second increments

Happy to implement these myself (assuming I can figure out the dev flow...) if you're amenable.

### Your help is needed!

- [X] Yes I can help with this feature request!

bbilly1:
Sounds good! Makes it familiar to regular YouTube. I have been using browser based shortcuts, Firefox for example: https://support.mozilla.org/en-US/kb/html5-audio-and-video-firefox, but yeah, that doesn't support all options, just like play/pause...

In any case, please look into this. There is some help with setting up your dev environment in the [Contributing.md](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md), also don't hesitate to reach out on Discord if you get stuck.

bakkot:
Closing as fixed by https://github.com/tubearchivist/tubearchivist/pull/342.

PR #342, opened 2022-10-23
add keyboard shortcuts to video player

bakkot:
Fixes https://github.com/tubearchivist/tubearchivist/issues/341.

Specifically, this adds:
- 'c' to toggle captions
- 'm' to toggle mute
- '>'/'<' to increase/decrease speed (by units of .25, up to 3x)
- left-arrow/right-arrow to skip around by 5 seconds
- space to toggle play/pause even if the video element isn't selected
- '?' to show shortcuts

### DIFF
--- a/tubearchivist/home/templates/home/video.html
+++ b/tubearchivist/home/templates/home/video.html
@@ -2,7 +2,9 @@
 {% block content %}
 {% load static %}
 {% load humanize %}
-<div class="video-main"></div>
+<div class="video-main">
+    <div class="video-modal"><span class="video-modal-text"></span></div>
+</div>
 <div class="notifications" id="notifications"></div>
 <div class="sponsorblock" id="sponsorblock">
     {% if video.sponsorblock.is_enabled %}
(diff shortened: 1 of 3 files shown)


bbilly1:
Excellent, thanks for getting this up and running so fast. I think you implemented all the most useful shortcuts! 

Looking at this: https://support.google.com/youtube/answer/7631406?hl=en, I think having `.` and `,` could also be a good addition, you know, to find these easter eggs that you see flickering by sometimes? :-) What do you think?

bakkot:
I would've added `.` and `,`, but unfortunately I don't actually know how to advance a `video` element by a single frame, at least not without knowing the frame rate of the underlying video.

bbilly1:
So good, you even get a little notification box with what you have just clicked, amazing! :-)

Can we apply the same for the inplace player? You know when you click on the video thumbnail it opens the player right there. I think this needs also needs the modal wrapper? Then all the templates with playable videos, has a `player-wrapper` div where the player will go like here for example: https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/home/templates/home/home.html#L92.

Then the `createPlayer` function builds the player here: https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/static/script.js#L348.

bakkot:
> Can we apply the same for the inplace player?

Good catch, done.

bbilly1:
Thanks a lot for implementing that, that's an excellent addition!

PR #344, opened 2022-10-24
Add LDAP attribute mapping env variables.

BrianCArnold:
When using a default Samba DC LDAP instance, `uid` isn't used to hold the username, so for this, and other LDAP implementations, it's necessary to be able to specify which LDAP attributes are actually used for first name, last name, username, email, etc.

This doesn't change the default behavior, as it uses the current hardcoded values as the default values instead of requiring admins to specify it if they are upgrading to a version containing this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,8 +125,12 @@ You can configure LDAP with the following environment variables:
  - `TA_LDAP_DISABLE_CERT_CHECK` (ex: `true`) Set to anything besides empty string to disable certificate checking when connecting over LDAPS.
  - `TA_LDAP_BIND_DN` (ex: `uid=search-user,ou=users,dc=your-server`) DN of the user that is able to perform searches on your LDAP account.
  - `TA_LDAP_BIND_PASSWORD` (ex: `yoursecretpassword`) Password for the search user.
+ - `TA_LDAP_USER_ATTR_MAP_USERNAME` (default: `uid`) Bind attribute used to map LDAP user's username
+ - `TA_LDAP_USER_ATTR_MAP_PERSONALNAME` (default: `givenName`) Bind attribute used to match LDAP user's First Name/Personal Name.
+ - `TA_LDAP_USER_ATTR_MAP_SURNAME` (default: `sn`) Bind attribute used to match LDAP user's Last Name/Surname.
+ - `TA_LDAP_USER_ATTR_MAP_EMAIL` (default: `mail`) Bind attribute used to match LDAP user's EMail address
  - `TA_LDAP_USER_BASE` (ex: `ou=users,dc=your-server`) Search base for user filter.
- - `TA_LDAP_USER_FILTER` (ex: `(objectClass=user)`) Filter for valid users. Login usernames are automatically matched using `uid` and does not need to be specified in this filter.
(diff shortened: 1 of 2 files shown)


BrianCArnold:
I figured out what I did wrong, stupid mistakes because I'm not a python programmer. Everything appears to work as desired now.

bbilly1:
Thanks for your addition. @DanielBatteryStapler is our resident LDAP expert who created the original implementation. Can you take a loot at these changes?

Then you might have seen, the linter is unhappy with the formatting. Mostly about line length and whitespaces. In the deploy.sh script there is a [validation](https://github.com/tubearchivist/tubearchivist/blob/master/deploy.sh#L75) function so you can run the linter locally.

BrianCArnold:
I see what the linter trying to get at. I wonder why it didn't have an issue with the original version, seems similar. I'm about to start my workday, I'll get back to it and fix it some time later today.

DanielBatteryStapler:
Changes look great to me, I approve. Thank you for your additions! I thought about including exactly this in my initial LDAP additions, but I got lazy and thought that the defaults would be good enough -- guess I was wrong about that one. Glad to see that get fixed.

BrianCArnold:
This should be fixed according to the linter.

BrianCArnold:
That's odd, the linter accepted it on my end...

bbilly1:
Almost there, flake8 doesn't like the long lines in settings.py between line 108 and 115. Unfortunately I need to run the action here manually for fist time contributors, it's a github thing...

BrianCArnold:
oh, i was pretty sure I fixed those too, I'll check.

bbilly1:
Python also has this nifty multi line comment feature, basically wrap it in `"""` like this:
```python
"""
this is a multi
line comment
spanning over
multiple lines,
nicely.
"""
```
Does the same as your solution, but you look more like a Python pro. :-)

DanielBatteryStapler:
I'm just gonna pop in to be annoying and say those aren't multi-line comments, those are multi-line string literals and you're simply ignoring their value! Still useful as comments though lol

BrianCArnold:
... I'm not sure if I love or hate that.

BrianCArnold:
The language behavior, not you popping in. I enjoy "technically correct", I am a programmer, a after all.

BrianCArnold:
Shortened line lengths.

BrianCArnold:
I officially renounce Python, lol

bbilly1:
Sorry, I took the liberty to fix the remaining few whitespace issues. I know, python linting needs some time to get used to. Please don't be discouraged, there are lots of great opportunities to help with this project! :-)

BrianCArnold:
I just need to take the time to actually get a decent python environment set up, i've been using a docker container to run black, but it's not ideal, and VSCode doesn't seem to be getting along with my Python environment.

bbilly1:
Yeah, after some time fixing the same issues over and over, you'll get the hang of it. I know it's the equivalent of a grammar naz*, but it does make the code much more readable, particularly if you have a few people working on the same code base...

Anyways, thanks for your contribution, all help is welcome with this project!

BrianCArnold:
I'm actually a huge stickler for linting and keeping code clean, it's just that I don't care as just about EOL white space, and I have my line length set as warnings.

PR #345, opened 2022-10-24
Format/slightly modernize the JS

bakkot:
This adds configuration for eslint and prettier, the de-facto standard linter and formatter (respectively) for JavaScript, and runs them on the codebase.

I've gone with very standard configuration for both. A lot of projects get more opinionated about all the various eslint rules, but in this PR I've kept it to what I consider to be the bare minimum for preventing errors.

I haven't set up automatic enforcement of these rules, because that would require users to have another tool (namely `node`) installed. I'm happy to do so if you'd like, though.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,12 @@
+module.exports = {
+  extends: [
+    'eslint:recommended',
+    'eslint-config-prettier'
+  ],
+  parserOptions: {
+    ecmaVersion: 2020,
+  },
+  env: {
+    browser: true,
+  },
+};
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,4 @@
+'use strict';
 module.exports = {
   extends: [
     'eslint:recommended',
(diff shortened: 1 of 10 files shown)


bakkot:
I've updated the `codespell` command in both the github action and in the `deploy.sh` script (thanks `@lamusmaser`!). It might make sense to instead change the github action to invoke `./deploy.sh validate`, instead of having to keep the logic in two places. But this works for the moment.

bbilly1:
hehe, lot's of changes. :-) But I think I understand everything that's happening. Some quick testing looked like all functionality was still behaving the same. I think I also get what the linter is up to and this makes things much more readable.

So do you think we can set this up as a github action too?

In any case, thanks for your help, and please help to keep these dependencies updated.

lamusmaser:
I can get the Github Action setup, since it is associated with #346. I was mostly just requiring a validation of what exactly the script was doing, in comparison to the other, and see if I could get it working. I might be able to setup an Action on my repo and confirm (or setup a private repo and do the same) since I have access to several runners personally.

PR #347, opened 2022-10-25
Make Dockerfile throw error instead of silently failing to install ffmpeg

BrianCArnold:
Creating an image with the Dockerfile will always complete, however when TARGETPLATFORM is null or a unexpected value, the resulting image does not include ffmpeg, causing containers to fail to actually download videos. 

This changes the behavior of the Dockerfile to fail on the ffmpeg installation step if TARGETPLATFORM is not `linux/amd64` or `linux/arm64`, as well as emitting a STDERR message and exiting with a 1, which stops the build on that step with that error message.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,9 @@ RUN if [ "$TARGETPLATFORM" = "linux/amd64" ] ; then \
         rm ffmpeg.tar.xz \
     ; elif [ "$TARGETPLATFORM" = "linux/arm64" ] ; then \
         apt-get -y update && apt-get -y install --no-install-recommends ffmpeg && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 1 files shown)


BrianCArnold:
Nevermind, looks like someone already addressed the situation.

bbilly1:
Hehe, sorry, I did that more or less in parallel. I thought for most usecases, default should be `linux/amd64`, so if you just run the Dockerfile, that's what you'll get. Then I also took the opportunity to get the custom ffmpeg builds into arm64: https://github.com/tubearchivist/tubearchivist/commit/f8dccfcd15b169946a3ff5c74cb588ce9e111e73. 

But as you can see, there are lot's of opportunity to improve things! I'm usually also around on Discord.

PR #348, opened 2022-10-26
add github workflow to enforce linting of JS files

bakkot:
Per https://github.com/tubearchivist/tubearchivist/pull/345#issuecomment-1291422057

I didn't fold this into the `deploy.sh validate` script because that would mean every contributor would need to have `node` installed, even if they weren't working on the JS parts at all.


bbilly1:
Excellent, thanks for implementing that, let's see how this is doing it's thing. :-)

PR #351, opened 2022-10-26
Update Github Actions to use `deploy.sh` and remove individual calls for testing functions.

### LINKED ISSUE #346: [Bug]: Github Actions and `deploy.sh` can become out of sync and produce different results for local and online testing
### Latest and Greatest

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).

### Operating System

Not Applicable/Linux

## Describe the bug
As noted in https://github.com/tubearchivist/tubearchivist/pull/345#issuecomment-1289768021, there can be a dereference between `deploy.sh` and the Github Actions regarding tests (which means that local and PR testing can change).

This means that we need to update the Github Actions to call `deploy.sh` and reference the output for linting/testing results.

## Steps To Reproduce
Create a PR with a modified `deploy.sh`, which causes different results for testing locally.

## Expected behavior
Local and Actions testing should be maintained until either more sophisticated testing replaces them or a better alternative is developed to replace the calls in `deploy.sh`.

### Relevant log output

```shell
Reference the Actions Logs before https://github.com/tubearchivist/tubearchivist/pull/345#issuecomment-1289768021.
```

### Anything else?

_No response_

lamusmaser:
Call `deploy.sh validate` for linting checks. Resolves #346 when completed.

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -4,13 +4,13 @@ jobs:
   lint_python:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
       # note: this logic is duplicated in the `validate` function in ./deploy.sh
       # if you update this file, you should update that as well
       - run: pip install --upgrade pip wheel
       - run: pip install bandit black codespell flake8 flake8-bugbear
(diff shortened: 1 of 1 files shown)


lamusmaser:
I additionally incremented the versions for the Github Actions from Github. This will remove the deprecation messages that were coming through for each run.

bbilly1:
Excellent, one source of truth, thanks for implementing that!

Issue #189, opened 2022-03-08
[Feature Request] More intuitive way to adjust/limit amount of videos to be downloaded


suckerfish:
I think when someone first adds a subscription it should be very easy to tell TA that you want to get the entire channel or not _or_ give some nice controls about which videos exactly they will be getting when they subscribe. Maybe radio buttons or checkboxes or something "Download entire channel?" and "Specify date range for which videos you want" or "Get X number of recent videos?" 

The only way to do this currently is to adjust a global setting about how far back TA looks for **all** subscriptions.

bbilly1:
Yes, but I'd say this would better fit on the downloads page than on the subscribe page. these are different tasks that run in the background. 

Then on the subscribe page, there could be a mention about how many videos this will add next time you rescan.

BlipRanger:
Yeah, regarding this; what is the correct way to instruct TA to archive an entire channel? Just bumping the limit up really high?

PR #352, opened 2022-10-27
Add FAQ item about downloading vs subscribing

bakkot:
This confused me, and apparently I'm [not the only one](https://github.com/tubearchivist/tubearchivist/issues/189#issuecomment-1118802407), so I figure it's worth writing down somewhere more prominent than an old issue.

Happy to put this somewhere else if you'd prefer. Or feel free to just close this if you think it's sufficiently discoverable already and I just missed it.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -29,3 +29,6 @@ So Docker is the only supported installation method. If you don't have any exper
 
 ## 4. Finetuning Elasticsearch
 A minimal configuration of Elasticsearch (ES) is provided in the example docker-compose.yml file. ES is highly configurable and very interesting to learn more about. Refer to the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/index.html) if you want to get into it.
+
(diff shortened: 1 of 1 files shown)


bbilly1:
Definitely helpful to have that there, thanks for getting the ball rolling. Maybe mention, that these are not mutually exclusive, you can download the whole channel *and* subscribe to it to get new videos in the future? Something along the lines like [here](https://github.com/tubearchivist/tubearchivist/issues/189#issuecomment-1118802407).

How I implemented that in the very beginning of this project made sense for me, but it is definitely opinionated how I use this project, but yeah, it has come up a lot...

bakkot:
> Maybe mention, that these are not mutually exclusive, you can download the whole channel and subscribe to it to get new videos in the future?

[Done](https://github.com/tubearchivist/tubearchivist/pull/352/commits/d09aef14020ab4af89dc92a81aeee79a5b2dd860).

bbilly1:
That's it, thanks, great addition!

PR #353, opened 2022-10-28
README updated with format update

lamusmaser:
Based on https://discord.com/channels/920056098122248193/1035107198998691890/1035141012600930414, adjusted the README to reflect an updated format string that can be used. This updates the Getting Started section of the README to include the adjustments necessary to actually download videos.


bbilly1:
Thanks, thanks a lot for implementing that so quickly. The same example is also on the settings page within Tube Archivist, please change it there too, so it will go into the next release: https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/home/templates/home/settings.html#L80

lamusmaser:
> Thanks, thanks a lot for implementing that so quickly. The same example is also on the settings page within Tube Archivist, please change it there too, so it will go into the next release: https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/home/templates/home/settings.html#L80

Done.

bbilly1:
Excellent, thanks a lot!

PR #357, opened 2022-11-01
Changed the wording in a few places

pairofcrocs:
Just updated some wording on the /channel and /downloads page.


bbilly1:
So the thinking here was, that depending if you have toggled the *show subscribed only* or *Show only ignored videos* that total will change. So total channels will be *total channels you are subscribed to* and not total channels indexed. Hence the *matching* there. Do you think that is obvious enough without specifying that?

pairofcrocs:
My thought was, if you click the toggle, you already know that it's going to hide them in that selection. Putting `matching` seemed redundant.

PR #367, opened 2022-11-21
Add podman installation instructions to docs

redxtech:
As mentioned in the discord server, I've added some additional information required to run Tube Archivist with podman.

I've also added a notice that podman users should check out the `docs/Installtion.md` page to make the changes necessary to get it to work.

Potential feedback spots:
 - Service file creation: while not strictly necessary, it brings the experience on-par with docker. If you don't want to link out to an external website, I can either inline the resulting service file, change to link to an archived version of the page, or I can remove the section altogether
 - Resulting `docker-compose.yml`: I can inline the included compose file with the mentioned changes, if deemed necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Once your YouTube video collection grows, it becomes hard to search and find a s
 - [Tube Archivist Metrics](https://github.com/tubearchivist/tubearchivist-metrics) to create statistics in Prometheus/OpenMetrics format.
 
 ## Installing and updating
-There's dedicated user-contributed install steps under [docs/Installation.md](./docs/Installation.md) for Unraid, Truenas and Synology which you can use instead of this section if you happen to be using one of those. Otherwise, continue on.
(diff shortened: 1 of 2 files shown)


bbilly1:
Excellent, thanks for working on that. I think your instructions are very clear and easy to follow. Also future proof, which is always a plus. The service file link could be helpful, and looks like the blog is a well established and reputable resource, so I think it's fine to have that there.

For the update in the readme, maybe instead of adding a separate paragraph, it would fit better with the other mentions: https://github.com/tubearchivist/tubearchivist/blob/master/README.md?plain=1#L64, so we only need one link to the installation.md file, and it's right there on top and if somebody is searching Ctrl+f for `podman`, they will find it immediately anyways.

redxtech:
That's a much better way of doing it, no need to have an extra paragraph where a single word works just fine!

I've made that change just now.

Issue #412, opened 2023-01-19
[Bug]: v0.3.1+ Single Download does not apply the correct `vid_type` when pulling from the ElasticSearch queue


lamusmaser:
### I've read the documentation

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [X] I have read through the [wiki](https://github.com/tubearchivist/tubearchivist/wiki) and the [readme](https://github.com/tubearchivist/tubearchivist#installing-and-updating), particularly the [common errors](https://github.com/tubearchivist/tubearchivist#common-errors) section.

### Operating System

Unraid, Synology

## Describe the bug
When attempting to download a single video, the videos do not download due to a KeyError exception occurring. 

## Steps To Reproduce
Attempt to download a single video through the "Download Now" button on the Downloads Page.

## Expected behavior
Video should download instead of producing the exception.

### Relevant log output

```shell
erAoSrjDf6c: change status to priority
[2023-01-19 12:49:53,542: INFO/MainProcess] Task home.tasks.download_single[2078dc6a-9ee6-43a6-815d-168ff6bd203e] received
[2023-01-19 12:49:53,547: ERROR/ForkPoolWorker-4] Task home.tasks.download_single[2078dc6a-9ee6-43a6-815d-168ff6bd203e] raised unexpected: KeyError('vid_type')
Traceback (most recent call last):
  File "/root/.local/lib/python3.10/site-packages/celery/app/trace.py", line 451, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/root/.local/lib/python3.10/site-packages/celery/app/trace.py", line 734, in __protected_call__
    return self.run(*args, **kwargs)
  File "/app/home/tasks.py", line 117, in download_single
    "vid_type": pending_video["vid_type"],
KeyError: 'vid_type'
```

bbilly1:
That should have gotten set during the migrations here: https://github.com/tubearchivist/tubearchivist/blob/3602ec45604a492d708cdf710302ade7f008aef4/tubearchivist/home/apps.py#L135-L160

That's for the videos in the download queue during migration, as well as your existing videos.

I'm guessing the startup checks aren't running or didn't run properly. Manually delete the leftover lock if needed:
```
docker exec -it archivist-redis redis-cli DEL ta:startup_check
```
then restart the container and monitor the logs for errors in during startup.

lamusmaser:
[...]

PR #413, opened 2023-01-19
412 fix `vid_type` with downloads (single and bulk)

lamusmaser:


lamusmaser:
This is associated with resolving #412

lamusmaser:
This is being updated to process a default `vid_type` if one is not present.

bbilly1:
Thanks for working on that. I just undid the TA_VERSION there for an easy merge, so we don't need to do a rebase.

Issue #409, opened 2023-01-16
[Feature Request]: Ability to visit source URL for preview from download queue


SmartPhoneLover:
### Already implemented?

- [X] I have read through the [wiki](https://github.com/tubearchivist/tubearchivist/wiki).
- [X] I understand the [scope](https://github.com/tubearchivist/tubearchivist/wiki/FAQ) of this project and am aware of the [known limitations](https://github.com/tubearchivist/tubearchivist#known-limitations) and my idea is not already on the [roadmap](https://github.com/tubearchivist/tubearchivist#roadmap).

### Your Feature Request

I would find useful to be able to visit the source URL of the video (YouTube link) from the download queue list.

![45754674747](https://user-images.githubusercontent.com/42898749/212658771-b65d56f0-b752-480c-84db-4f8238fa3af4.JPG)

It could be the title of the video itself (clickable), like shown in the pic above. Or maybe, add another button below to visit the link.

I ask for it because, more than once I want to preview the video before downloading and I have to open YouTube page in another window and copy/paste the title to locate it. So, it would save time for this task.

What do you all think about it? Thank you.

### Your help is needed!

- [ ] Yes I can help with this feature request!

bbilly1:
Thanks for implementing this @deltacodepl, this is now published in v0.3.3.

PR #422, opened 2023-02-01
Ability to visit source URL for preview from download queue

deltacodepl:
Small change - just adds link to each YouTube videos on downloads page
#409


bbilly1:
Thanks a lot for implementing this so quickly. Please ignore the linter error, it's not you, it's me. :-)

I'll merge it in a bit, I've done some refactoring, but that shouldn't affect your change here.

PR #426, opened 2023-02-08
Add search usage summary on the search page and improve the search docs

p0358:
When search query is empty, it adds this page which summaries the basic search (along with some examples), so that the user can quickly grasp/look up how search is performed without reading up the wiki:
image

The docs are improved a tad too to *hopefully* make it more clear how search keywords work to the reader.

### DIFF
--- a/docs/Search.md
+++ b/docs/Search.md
@@ -1,31 +1,32 @@
 # Search Page
 Accessible at `/search/` of your **Tube Archivist**, search your archive for Videos, Channels and Playlists - or even full text search throughout your indexed subtitles.
 
+Just start typing to start a **simple** search *or* **start your query with a primary keyword** to search for a specific type and narrow down the result with secondary keywords. Secondary keywords can be in any order. Use *yes* or *no* for boolean values.
+
+- This will return 30 results per query, pagination is not implemented yet.
 - All your queries are case insensitive and are normalized to lowercase.
 - All your queries are analyzed for the english language, this means *singular*, *plural* and word variations like *-ing*, *-ed*, *-able* etc are treated as synonyms.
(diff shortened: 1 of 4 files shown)


bbilly1:
Lot's of great stuff here. Wiki section looks great, I think that will make things much clearer for people. 

For the search.html changes, I think that's on the right track. Saw the `<tt>` html tags, that's a blast from the past, I think they are [depreciated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt), maybe use `<code>` instead? Then for the colors, better to use the variables from the CSS file, so it will adapt based on the current and maybe future color schemes. I think `accent-font-dark` and `highlight-bg` would work? Maybe add a very small padding of something like `3px` for a little better spacing?

For the basic search query, let's use something less controversial. :-) How about `python django`, at least for me that returns a lot of great videos.

Thanks a lot for your help on this one!

p0358:
Updated looks

image

I didn't use `<code>` because iirc from my previous attempts, that implies some CSS on its own, so it was easiest to just use `<span>` since `<tt>` only implied monospace font family at this point.

I'll also paste these messages from discord here for future reference:

> I wanted "basic search" example to be something outside of programming bubble to be also a very very slight joke about it being "basic" search

> and so I thought of gangam style first as one of the most viewed videos, first one to hit billion views, and probably the least hated of the videos from the top viewed list

> as for the change of `<tt>` tag, that's defo reasonable, and that tag is (supposed to be) only styled with font-family: monospace; either way

> but as for padding I wasn't sure either

> because the queries are without spaces, and otherwise it could give people bad idea, so I thought differentiating with colors would be a better idea, but then thought there might not be enough accent colors for that

> but maybe if I used the two shades of accent as background, it could work out

> I differentiated the background colors in hope to make it more readable

bbilly1:
Excellent work, thanks for your input. I've fine tuned the colors a bit to make sure foreground and background remain readable for different color themes.

Issue #415, opened 2023-01-22
[Blocking]: List of things to do before working on new features


bbilly1:
Background: This project has grown very fast, too fast, new features were implemented at the cost of reduced maintainability and reliability of the project. Functionality wise, this project has a lot going for it, now its time to revisit some of these early flaws, so:

This is a potentially none complete list of blocking things to figure out, before working on additional major features. Outlined solutions may or may not be good ideas.

- [X] Fix startup process and checks
  - Create manage.py function to check environment
  - Create manage.py function to apply custom migrations
  - All single threaded, no locks needed, track exit status to stop startup on errors
- [X] Fail prove process locks
  - Track process ids produced by Celery to be able to check if they are running independent from locks
- [X] Enforce snapshots
  - Guarantee snapshots are configured and enabled on startup, disable json backup task
- [X] Complete search processor
  - Finalize rewrite serialize classes for dynamically process values from ES
  - Move all functionality to API
- [X] Complete API
  - Move all form and buttons POST request to API endpoints
- [X] Split up settings page into separate sub pages
- [x] Transfer queue storage to redis
  - Load all tasks queues into redis for celery to take from one by one
- [x] Rewrite cron schedule validator
  - Use Celery to validate input first before storing
- [X] Refactor Subscribe functionality
  - Make distinction between Playlist and Channel
  - Surface error notifications through new task backend
- [X] Refactor get next item to download
  - Call get next on the ES queue to get whole video context
  - Use new task backend to interact with the queue
  - Implement priorities and auto download in ES
- [X] Simplify media path building
  - Standardize media paths to `/<channel-id>/<video-id>.mp4`
  - Fix all these incompatibilities in one go
- [ ] Remove all usecase of Redis plugins
  - That has broken too many times, it's time to just stick to standard redis
  - make all redis data transient, use only ES for persistence. 

<hr>

Additionally, not strictly blocking but high priority:
- [x] At least a minimal automatic testing suite that can be extended
- [ ] Create useful content for https://www.tubearchivist.com
- [ ] Prepare project for a *All in one* version:
  - Figure out how to pack this as efficiently as possible
  - Figure out automated build process
  - Make container internal base paths configurable
  - Make ES auth optional
- [X] Implement the notification broker on https://members.tubearchivist.com to collect webhooks from YouTube for realtime subscription notifications. 

<hr>

This list will be updated as we go. Expect this to take as long as it takes. Any help with this is highly welcome. Reach out on Discord to coordinate.

[...]

PR #427, opened 2023-02-08
Refresh progress on downloads page more often (3000 ms => 500 ms)

p0358:
I believe it makes sense to do that, because 3 seconds is quite a slow rate. And most people sitting on Downloads screen, I assume, are watching as their downloads are getting done, and could appreciate snappier refresh of the progress. I think 500 ms is still a rather conservative time, and should not pose increased strain on the server, given it just makes a few redis get queries to grab it. In any case, even if network or server is overloaded, the requests are only issued 500 ms *after* previous response is received, and not just on constant interval irregardless of whether previous response arrived.

With all the above considered, I think the (negligible) const of refreshing the progress text a bit often is worth it for the sake of nicer/snappier/more up-to-date progress display, especially that the requests only keep getting issued while any videos are being downloaded and stop once progress becomes empty.

[ Of course for the snappiest experience it would be ideal to implement a websocket endpoint which could push progress as it happens, but that would require substantially more work to code and ensure it works properly... (Redis has its keyspace notifications that could potentially suit there, but not sure if the Python framework in use even supports websocket endpoints in the first place + also reverse proxies would need to support it fine or we'd need to have a fallback). Alternatively server-sent events could be considered, and perhaps that would be way easier than websockets. ]

### DIFF
--- a/tubearchivist/static/progress.js
+++ b/tubearchivist/static/progress.js
@@ -28,16 +28,12 @@ function checkMessages() {
 // get messages for page on timer
 function getMessages(dataOrigin) {
   fetch('/progress/')
-    .then(response => {
-      return response.json();
-    })
+    .then(response => response.json())
     .then(responseData => {
-      let messages = buildMessage(responseData, dataOrigin);
+      const messages = buildMessage(responseData, dataOrigin);
       if (messages.length > 0) {
         // restart itself
-        setTimeout(function () {
-          getMessages(dataOrigin);
-        }, 3000);
+        setTimeout(() => getMessages(dataOrigin), 500);
       }
     });
 }


p0358:
But speaking of server-sent events, I made a prototype commit there: https://github.com/p0358/tubearchivist/tree/feature-sse-progress ( https://github.com/p0358/tubearchivist/commit/214c2440ee80b3a4c3a14049077a95e570fca6a0 )
which did client-side part and mock of server-side.
Proper server-side requires using some form of pub/sub (with keyspace notifications for instance), I left the comments for that in the .py file (as frankly I don't really know how to do what I described there in Python...):
```py
def progress(request):
    # pylint: disable=unused-argument
    """resolves to /progress/
    return list of messages for frontend
    """
    if request.GET.get("stream", False):
        # TODO: use proper pub/sub callbacks instead of this
        # hint: https://redis.io/docs/manual/keyspace-notifications/
        # https://medium.com/@imamramadhanns/working-with-redis-keyspace-notifications-x-python-c5c6847368a
        # - we probably want to use keyspace events, and thus `Kg$d` config (to catch JSON and string key types operations)
        # - keyspace notifications need to be enabled, at runtime should be doable with CONFIG SET
        # - we probably need some common place where it'd SUBSCRIBE to all possible keys in the main TA app, and then
        # the generator progress_event_stream() should somehow add itself as one of the callbacks and later remove
        # (but how do we know if the request was terminated BY CLIENT mid-way to do that?)
        # TODO: read request.GET.get("types") to listen just to specified types of interest
        def progress_event_stream():
            all_messages = RedisArchivist().get_progress()
            yield json.dumps({"messages": all_messages})
            while len(all_messages):
                sleep(0.5)
                all_messages = RedisArchivist().get_progress()
                yield json.dumps({"messages": all_messages})
        return StreamingHttpResponse(progress_event_stream(), content_type="text/event-stream")
    else:
        all_messages = RedisArchivist().get_progress()
        json_data = {"messages": all_messages}
        return JsonResponse(json_data)
```

bbilly1:
Yeah, your reasoning is sound. I think folks who access their TA over LAN will appreciate it. If you have higher latency because you access your homelab from outside your network over a tunnel or something, you might not get 500ms between responses, yeah, that was the reasoning to not do that asynchronously, so wait for the response before setting the timeout for the next request. 

I've been meaning to parse the progress to make a little progress bar inside the message box... At some point. :-)

Websockets, I know, that's how you'd do it. I've looked into it before, as you mentioned, that would require substantial refactoring, we'd have to ditch uwsgi for an asgi capable server, django has [django channels](https://channels.readthedocs.io/en/stable/) to handle these things quite conveniently. But I'm not quite ready to tackle that, want to make progress on #415 first.

PR #445, opened 2023-03-05
using docker networking instead of exposing ports

gautamkrishnar:
This change will prevent Elasticsearch and Redis from being exposed to the public internet when deployed on cloud environments. Uses docker networking instead.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,8 @@ services:
     image: bbilly1/tubearchivist
     ports:
       - 8000:8000
+    networks:
+      tubearchivist-net
     volumes:
       - media:/youtube
       - cache:/cache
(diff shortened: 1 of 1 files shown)


lamusmaser:
Expose leaves it internal and therefore does not expose it to the Internet for public-facing environments, [as noted in the docker-compose documentation](https://docs.docker.com/compose/compose-file/compose-file-v3/#expose). If started via docker-compose, then a non-prescribed network will actually create a network just for the stack elements to communicate. This is also noted in the [Docker documentation](https://docs.docker.com/compose/networking/).

If someone wants to generate their own network, then that is fine. However, there is a percentage of the population that run multiple instances of TubeArchivist, for various reasons, and including this as part of the base `docker-compose.yml` would cause conflicts for users that may not be as aware of the networking specifics of Docker.

gautamkrishnar:
Thanks for the explanation. Sounds good. Closing this one. 👍🏻

PR #446, opened 2023-03-06
reset token api refactor

GavinPalmer1984:


bbilly1:
Nice, that looks good. I think the only thing missing is a mention in the API docs here: https://github.com/tubearchivist/tubearchivist/tree/master/tubearchivist/api, maybe under *additional*? 

Also the linter likes an additional line there above the `TokenView` class definition. 

So for the dev environment setup, looks very interesting, hooking that up directly into VS code. The problem I see is maintainability, having two sets of Dockerfiles, docker compose ymls, run scripts, we'll always have to remember to make changes at both places. So far we don't really have a standardized dev environment over all contributors, because everybody has their own workflow, using what works for them.

So over all, this depends on your level of commitment, if you are on board with the project and you are up for keeping things consistent, I'm happy to merge and I think other might find this useful. Otherwise maybe an more abstract mention in the CONTRIBUTING file https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#development-environment might help as well?

PR #458, opened 2023-03-27
Fix video skipping 1% instead of 5s

Sungray:
When the video is focused, the built-in behavior is to skip 1% of the video instead of 5s, which isn't good.

### DIFF
--- a/tubearchivist/static/script.js
+++ b/tubearchivist/static/script.js
@@ -1458,19 +1458,11 @@ function doShortcut(e) {
       break;
     }
     case 'ArrowLeft': {
-      if (targetName === 'video') {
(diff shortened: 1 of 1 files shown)


bbilly1:
@bakkot generously implemented that feature in #342. Maybe you have some insights? Did you encounter any downsides with applying the same skip functionality in full screen mode?

bakkot:
I'm pretty sure this won't actually stop the browser controls from taking effect; it'll just apply both skips (so 1% + 5s) - probably not what you want. I can submit a different fix though.

bbilly1:
Thanks for all your help looking into this. I'll merge #459.

Issue #316, opened 2022-08-22
[Feature Request]: Implement some additional yt-dlp options/commands


M4NH3X:
### Already implemented?

- [X] I have read through the [wiki](https://github.com/tubearchivist/tubearchivist/wiki).
- [X] I understand the [scope](https://github.com/tubearchivist/tubearchivist/wiki/FAQ) of this project and am aware of the [known limitations](https://github.com/tubearchivist/tubearchivist#known-limitations).

### Your Feature Request

Currently we can't pass custom yt-dlp commands, it would be nice to have that option.

I know if we allow all options then it will break TA but yt-dlp is really robust, so that we can allow few whitelisted commands so that things won't break.

Thank you for this great project!

------

If this request is rejected for some reason, would love to see these options integrated into TA web-ui themselves. 

* [Sorting Formats:](https://github.com/yt-dlp/yt-dlp#sorting-formats) `-S` (this is far more robust than -f)
* Extract audio: `-x` `--audio-format` (Channels dedicated to podcasts, etc. can be really helpful)
* `-N` or `--concurrent-fragments` (some formats use fragments)
* `--write-info-json` `--write-description` (it will increase portability of downloaded data if users plans to migrate)

That's all options I can think of now which won't break things but increase flexibility.

### Your help is needed!

- [ ] Yes I can help with this feature request!

bbilly1:
So with arbitrary settings import it's going to be very easy to break Tube Archivist. As that is in essence untestable. So it's not a question if yt-dlp is robust, which it is, it's that Tube Archivist expects certain files to be there, and other not to be there. Also depending on the option you choose, you'll want to index things differently, but that wouldn't be possible with arbitrary settings. 

I guess that's the difference between a download GUI that only needs to worry about that once at download, and a media server where everything needs to fit with each other indefinitely. 

But for your configuration options:
1. Sorting formats: Sounds good, that's only at download time, so that's straight forward to implement.
2. Extract audio: That's on the roadmap with the podcast rss feed, so that can be implemented together with that.
3. `-N`: I'm not familiar with that one, but sounds like a download only option? Need to look into that.
4. Write description is implemented with `--embed-metadata`, which also embeds description plus more. I think there is also `--embed-info-json`? Does that work with mp4 files? I'd rather not create separate files for that, as this will require more changes to handle that for things like *rescan filesystem* and the like and will double the amount of files you create, that's why you have a media server! Plus embedded things are easy enough to extract if portability is your concern. 

[...]

PR #471, opened 2023-04-27
Allow to configure yt-dlp `--format-sort` argument

dsander:
This exposes the [`--format-sort`][1] yt-dlp option to the user.

Implements parts of #316

[1]: https://github.com/yt-dlp/yt-dlp#sorting-formats

### DIFF
--- a/tubearchivist/home/src/download/yt_dlp_handler.py
+++ b/tubearchivist/home/src/download/yt_dlp_handler.py
@@ -312,6 +312,8 @@ class VideoDownloader:
         """build user customized options"""
         if self.config["downloads"]["format"]:
             self.obs["format"] = self.config["downloads"]["format"]
+        if self.config["downloads"]["format_sort"]:
+            self.obs["format_sort"] = self.config["downloads"]["format_sort"].split(",")
         if self.config["downloads"]["limit_speed"]:
             self.obs["ratelimit"] = (
                 self.config["downloads"]["limit_speed"] * 1024
--- a/tubearchivist/home/src/download/yt_dlp_handler.py
+++ b/tubearchivist/home/src/download/yt_dlp_handler.py
@@ -313,7 +313,9 @@ class VideoDownloader:
         if self.config["downloads"]["format"]:
             self.obs["format"] = self.config["downloads"]["format"]
         if self.config["downloads"]["format_sort"]:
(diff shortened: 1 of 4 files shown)


bbilly1:
Excellent, thanks for taking this on! :-)

You can register a default value in the config template [here](https://github.com/tubearchivist/tubearchivist/blob/testing/tubearchivist/home/config.json), then there is a startup management function that will load that into redis [here](https://github.com/tubearchivist/tubearchivist/blob/testing/tubearchivist/config/management/commands/ta_startup.py#L48-L57). This way we can assume everybody has that key set in their config, even if it's just set to `False` and we don't have to worry about any KeyErrors or such.

Then it might be worth it to mention in the markup on the settings page that no all codecs are compatible with all browsers. Especially Safari is very picky. Just in case people get confused why their video doesn't have audio anymore.

tubearchivist/home/src/download/yt_dlp_handler.py
```diff
@@ -312,6 +312,8 @@ def _build_obs_user(self):
         """build user customized options"""
         if self.config["downloads"]["format"]:
             self.obs["format"] = self.config["downloads"]["format"]
+        if self.config["downloads"]["format_sort"]:
+            self.obs["format_sort"] = self.config["downloads"]["format_sort"].split(",")
```
bbilly1, line 316: I think yt-dlp is picky here, I think they expect no spaces after the commas. But we humans have that kind of ingrained. Would be a nice quality of life if we can handle both here, something like:
```python
format_sort_list = [i.strip() for i in format_sort.split(",")]
```

dsander: Good point, done. I hope the style is OK.

dsander:
> You can register a default value in the config template [here](https://github.com/tubearchivist/tubearchivist/blob/testing/tubearchivist/home/config.json), then there is a startup management function that will load that into redis [here](https://github.com/tubearchivist/tubearchivist/blob/testing/tubearchivist/config/management/commands/ta_startup.py#L48-L57). This way we can assume everybody has that key set in their config, even if it's just set to `False` and we don't have to worry about any KeyErrors or such.

Thanks for the hint, I missed the default config when grepping through the code base to figure out where this change might need to be made 😄

Issue #490, opened 2023-06-27
[Feature Request]: API endpoint for TA Version


lamusmaser:
### I've read the documentation

- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [feature request](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#feature-request) section.

## Is your feature request related to a problem? Please describe.
Unable to determine TA version through the API currently.

## Describe the solution you'd like
Develop an endpoint that provides the current TA version as a response.
Update the ping/pong endpoint to provide the current TA version as an addition JSON element.

## Additional context
None required, discussed on Discord through the `#help-contribute` channel.

### Your help is needed!

- [X] Yes I will work on this in the next few days or weeks.

lamusmaser:
Included an update to the docs: https://github.com/tubearchivist/docs/pull/13

No additional action required for feature at this time.

PR #491, opened 2023-06-27
Implement #490 - Add version API and add local_version function.

lamusmaser:
Thank you for taking the time to improve this project. Please take a look at the [How to make a Pull Request](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-make-a-pull-request) section to help get your contribution merged.

You can delete this text before submitting.

### DIFF
--- a/tubearchivist/api/urls.py
+++ b/tubearchivist/api/urls.py
@@ -5,6 +5,7 @@ from django.urls import path
 
 urlpatterns = [
     path("ping/", views.PingView.as_view(), name="ping"),
+    path("version/", views.VersionView.as_view(), name="api-ta-version")
     path("login/", views.LoginApiView.as_view(), name="api-login"),
     path(
         "video/",
--- a/tubearchivist/api/urls.py
+++ b/tubearchivist/api/urls.py
@@ -5,7 +5,7 @@ from django.urls import path
 
 urlpatterns = [
     path("ping/", views.PingView.as_view(), name="ping"),
-    path("version/", views.VersionView.as_view(), name="api-ta-version")
+    path("version/", views.VersionView.as_view(), name="api-ta-version"),
     path("login/", views.LoginApiView.as_view(), name="api-login"),
     path(
         "video/",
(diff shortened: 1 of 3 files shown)


lamusmaser:
Adds functionality for #490.

bbilly1:
Thanks for getting to work on that!

I see you opted for adding another endpoint? That also just returns the is_unstable? I think we can do that in the ping endpoint too, no? So we don't end up with duplicated functionality, two endpoints who do the same. Also the browser extension already uses the ping api everytime you open the popup, so we could see both there too, also if it's unstable build or not.

lamusmaser:
Fair enough. Looking at it closer, it looks like that does just parse it from the `TA_VERSION` variable anyway, so I will remove the items around the `is_unstable`. Beyond that, it would not add additional functionality, so I'll remove the additional endpoint for just the versioning.

bbilly1:
And you'll have to initiate the class first before you can call a method on it, e.g. like:
```
ReleaseVersion().get_local_version()
```

Otherwise you'll get `missing 1 required positional argument: 'self'` because it's an unbound method call.

lamusmaser:
I thought I had done it, but I must have missed it when I was moving the function calls around earlier, before committing. Good catch!

Anything else?

bbilly1:
Look good, please also update the docs, also with what that now returns, e.g.:
```json
{
    "response": "pong",
    "user": 1,
    "version": [
        0,
        3,
        6
    ]
}
```
btw: Do we want the parsed version like that? Makes sense if you need it to compare...

lamusmaser:
Was already planning to push docs PR tomorrow or Thursday with that update - I wanted to get the PR accepted first so I knew what I needed to include at the end.

I was going to go with the parsed only, since this is mostly meant to be programmatically interacted with long-term, therefore allowing for the easiest controls around version comparison. If you think it would be good, I could also add a `version_str` output, but I figured this was best for the immediate use cases (extension, media player plugins, quick interactions).

PR #506, opened 2023-07-31
updated imagry

pairofcrocs:
Updated screenshots, and welcome banner.


bbilly1:
very cool the mockup! That's what you get when you have professionals doing this! :-)

I was thinking, if you could add a search page screenshot too to showcase.md? As a central feature of this project, would be nice if you can see it right there with what you can do.

Maybe at the bottom of showcase the login page? That will be the first thing people see when they install the project, so they have already seen what to expect in the screenshots.

bbilly1:
ah, and maybe keep [tube-archivist-banner.jpg](https://github.com/tubearchivist/tubearchivist/blob/master/assets/tube-archivist-banner.jpg) in the repo there, who knows, maybe somebody wants to use that for their dashboard or something...

pairofcrocs:
[tube-archivist-social-preview.jpg](https://github.com/tubearchivist/tubearchivist/blob/master/assets/tube-archivist-social-preview.jpg) exists. I just thought it was a bit redundant to have both.

bbilly1:
yeah, good point. Folks can crop that if need be...

PR #510, opened 2023-08-06
Add `reindex` notifications category to Downloads tab

p0358:
It is somewhat weird that this particular progress seems to only be shown on a single channel view (while it's not channel-specific), and in the settings page.

I think it would make sense and be intuitive to show reindex progress on the Downloads tab. After all, you do see indexing progress there if you add up videos there manually, which is the first thing users do when trying out TA for the first time, so they grow an expectation to see the indexing-related progresses there in general I'd say...


bbilly1:
as discussed on discord, we will revisit that in favor of a more comprehensive notification overview.

Issue #457, opened 2023-03-26
[Bug]: Scheduler config doesn’t persist


s-h-a-d-o-w:
### I've read the documentation

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [X] I have read through the [wiki](https://github.com/tubearchivist/tubearchivist/wiki) and the [readme](https://github.com/tubearchivist/tubearchivist#installing-and-updating), particularly the [common errors](https://github.com/tubearchivist/tubearchivist#common-errors) section.

### Operating System

Ubuntu server 22

## Describe the bug

Scheduler config doesn’t persist.

## Steps To Reproduce

- Set a schedule for scanning and/or downloading in the web UI
- Refresh the page, settings should still be listed as you entered them
- Run `docker compose down` and `docker compose up`
- Go back to the settings - scanning and downloading schedules will now say `False` again

## Expected behavior

Config should be retained.

### Relevant log output

```shell
I’m not working on this project and there is a ton of log output. If you can be a bit specific what to look for, I’d be happy to check though.
```

PhuriousGeorge:
I'm going to start things out simple.  Did you scroll down and click the "Update Scheduler Settings" button after entering your values?  Once you click that, the page should refresh with your current settings.

These settings are stored in Redis, so unless there was a connection issue with your Redis setup or you don't have a persistent data directory setup for Redis appropriately.

s-h-a-d-o-w:
> I'm going to start things out simple. Did you scroll down and click the "Update Scheduler Settings" button after entering your values? 

Whoops, yes. Forgot that step in my OP, will update.

> These settings are stored in Redis, so unless there was a connection issue with your Redis setup or you don't have a persistent data directory setup for Redis appropriately.

I probably also should have been more specific about the fact that I'm using the docker compose workflow, for which no redis setup is required based on the docs.

PhuriousGeorge:
> I probably also should have been more specific about the fact that I'm using the docker compose workflow, for which no redis setup is required based on the docs.

I'm uncertain where you're assuming Redis isn't required.  See relevant section of the example [compose file](https://github.com/tubearchivist/tubearchivist/blob/master/docker-compose.yml#L26)

s-h-a-d-o-w:
Unless I'm missing something, the containers are created based on the definition in the compose file, no additional work on behalf of the user required.

[...]

Issue #517, opened 2023-08-15
[Bug]: redis update can't read old dump.rdb file


Aeristoka:
### I've read the documentation

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

Docker

### Your Bug Report

https://registry.hub.docker.com/r/redis/redis-stack-server/tags

Redis 7.2.0-v0 Released today under the "latest" tag, redis failed to start, and thus TubeArchivist would not start, waiting on Redis

### Relevant log output

```shell
=~=~=~=~=~=~=~=~=~=~=~= PuTTY log 2023.08.15 17:32:38 =~=~=~=~=~=~=~=~=~=~=~=
docker logs archivist-redis 
[?2004l
8:C 15 Aug 2023 23:32:13.180 # WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. Being disabled, it can also cause failures without low memory condition, see https://github.com/jemalloc/jemalloc/issues/1328. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
8:C 15 Aug 2023 23:32:13.180 * oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
8:C 15 Aug 2023 23:32:13.180 * Redis version=7.2.0, bits=64, commit=00000000, modified=0, pid=8, just started
8:C 15 Aug 2023 23:32:13.180 * Configuration loaded
8:M 15 Aug 2023 23:32:13.180 * monotonic clock: POSIX clock_gettime
8:M 15 Aug 2023 23:32:13.180 * Running mode=standalone, port=6379.
8:M 15 Aug 2023 23:32:13.182 * <search> Redis version found by RedisSearch : 7.2.0 - oss
8:M 15 Aug 2023 23:32:13.182 * <search> RediSearch version 2.8.4 (Git=HEAD-7797d39d)
8:M 15 Aug 2023 23:32:13.182 * <search> Low level api version 1 initialized successfully
[...]

PR #521, opened 2023-08-20
change redis persistence from 1 hour (default) to 1 minute

contributor:
Update docker-compose.yaml sample file to configure redis persistence frequency

Fix https://github.com/tubearchivist/tubearchivist/issues/457

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,6 +31,8 @@ services:
       - "6379"
     volumes:
       - redis:/data
+    environment:
+      - REDIS_ARGS=--save 60 1              # if at least 1 key is changed - perisist data to volume after 60 seconds
     depends_on:
       - archivist-es
   archivist-es:


bbilly1:
Did some research on that, I hear you and I see I was under the wrong assumption that redis triggers a write automatically on shutdown.

Ultimately in the long run, we will want to move away from redis persistence, that has been more trouble that it's worth: #517. But that's unclear on what time frame that will happen.

Ultimately changing the compose file is tricky and manual on user side, unclear how many that will actually do. Also whatever you set the schedule to, 1min might not be fast enough, e.g. save config and restart immediately wouldn't do it, plus that will create a lot of unnecessary writes. 

Alternatively how about we call [BGSAVE](https://redis.io/commands/bgsave/) application internally whenever we write a persistent value to redis, basically if not expire [here](https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/home/src/ta/ta_redis.py#L38-L55). So it's tied to the application logic, only called when needed, folks don't need to update their compose files, and easy enough to revert once we can avoid persistence altogether in the future. 

Thoughts?

bbilly1:
OK, we are not the first ones to get confused here: https://github.com/redis-stack/redis-stack/issues/371

That's apparently an desired difference with the main redis image, ans also with our previous RedisJSON image.

contributor:
BGSAVE sounds good. But not sure if this worth efforts to call it all places from the app code if the plan to get rid of redis. And for SIGTERM, BGSAVE will only work if redis containter stops after TA container.

Still I believe additional line in compose is easy work-arround for time being, to prevent new users (like me) to give up on TA. I almost did that after spending many hours to find out why settings are never saved. I'm glad I didn't give up - the app looks fantastic (I'm mainly interested in comments/subtitle indexing feature)

contributor:
Good news. What about queue and task? Don't they need bgsave?

bbilly1:
queue and tasks in redis are adhoc anyways. they are usually built from persistent ES and loaded into redis for the duration of the task to take from. but they don't survive a container restart anyways and can always be recreated from ES again.

PR #524, opened 2023-08-23
Add fullscreen hotkey to video player

dmynerd78:
When browsing videos online, I often use the `f` hotkey to toggle fullscreen. It seems like the native player in Chrome and Firefox didn't support this, so I attempted to add it.

A few things I'm unsure about:
- This is the first time I'm using the fullscreen API and I may have missed some gotchas
- [`document.exitfullScreen`](https://caniuse.com/?search=document.exitfullScreen) and [`requestFullscreen`](https://caniuse.com/?search=requestFullscreen) are only supported by 80% of devices
- According to the caniuse above, this solution will not work on iphones. I'm not sure if there's a solution as I don't have an iphone to test with

### DIFF
--- a/tubearchivist/static/script.js
+++ b/tubearchivist/static/script.js
@@ -1483,6 +1483,23 @@ function doShortcut(e) {
       player.muted = !player.muted;
       break;
     }
+    case 'f': {
+      e.preventDefault();
+      if(document.fullscreenElement === null) {
+        player.requestFullscreen()
+        .catch(e => {
+          console.error(e);
+          showModal('Unable to enter fullscreen', 3000);
+        });
+      } else {
+        document.exitFullscreen()
+        .catch(e => {
+          console.error(e);
+          showModal('Unable to exit fullscreen', 3000);
+        });
+      }
+      break;
+    }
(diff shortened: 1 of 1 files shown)


bbilly1:
nice, very convenient. Tested on firefox and chromium. If a device doesn't support the JS methods, there is still always the fullscreen button from the player itself.

@bakkot initially implemented that, any thoughts?

bakkot:
Seems fine.

iPhones aren't going to be using the keyboard shortcuts anyway, so I wouldn't worry about that.

dmynerd78:
Doh! Yeah mobile would generally use the UI buttons over a hotkey I don't know why I got stuck on that.

If there's no other suggestions I'd be ready to merge.

bbilly1:
Very nice! If you have the time, please also open an PR to the docs to add it to the list of keys there, so we can just merge it at release.

dmynerd78:
Here's the PR for the docs https://github.com/tubearchivist/docs/pull/17. This is my first real time doing PRs so hopefully I didn't do anything incorrectly.

PR #531, opened 2023-09-01
Document how to update in the readme

bakkot:
I went to the `Update` section of the readme expecting to find a description of how to update, but it wasn't there. Found it in the discord instead and figured I'd put it here for any future readers.

If there's a different update process, happy to change the readme so it describes that instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Take a look at the example [docker-compose.yml](https://github.com/tubearchivist
 ## Update
 Always use the *latest* (the default) or a named semantic version tag for the docker images. The *unstable* tags are only for your testing environment, there might not be an update path for these testing builds. 
 
-You will see the current version number of **Tube Archivist** in the footer of the interface. There is a daily version check task querying tubearchivist.com, notifying you of any new releases in the footer. To take advantage of the latest fixes and improvements, make sure you are running the *latest and greatest*. After updating, check the footer to verify you are running the expected version.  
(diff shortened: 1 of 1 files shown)


bbilly1:
Ultimately that depends on what platform you are using. I think around 50% of TA users use regular docker compose. Then the rest use any of the myriad of other ways that you can run a docker container and they have their own ways how to update.

But ultimately, there aren't any special steps (well besides reading the release notes), you can update TA the same way as you update any docker container, pull and rebuild, that's not specific to TA but that will also be different between platforms.

So I'd say to keep the instructions in the readme generic, so they apply to all/most, then we still have the dedicated instructions in the docs. Maybe that would fit there better?

bakkot:
> But ultimately, there aren't any special steps (well besides reading the release notes), you can update TA the same way as you update any docker container, pull and rebuild, that's not specific to TA but that will also be different between platforms.

I can change it to say "for example", if that would help? I think anyone using something other than the basic setup is probably familiar with how to do updates on their platform. But I think it's worth having instructions for the basic setup in the readme. For some people - for example, me - this is the only docker-compose project they're using. I think it's worth having a sentence giving instructions for such people in the readme.

bbilly1:
Sounds good, maybe change it so something like *depending on your platform* and *for example*. Also there is no need for `down`, just `pull` and `up` should do it.

bakkot:
Done.

PR #533, opened 2023-09-02
Move user configuration from Redis to ES

anonamouslyginger:
Marking as draft as I have a bunch of testing of the migration to do but at least everyone can see what I'm up to

### DIFF
--- a/tubearchivist/home/src/es/index_mapping.json
+++ b/tubearchivist/home/src/es/index_mapping.json
@@ -1,5 +1,63 @@
 {
-    "index_config": [{
+    "index_config": [
+        {
+            "index_name": "users",
+            "expected_map": {
+                "colors": {
+                    "type": "text"
+                },
+                "page_size": {
+                    "type": "integer"
+                },
+                "sort_by": {
+                    "type": "text"
+                },
+                "sort_order": {
+                    "type": "text"
+                },
+                "view_style_home": {
+                    "type": "text"
+                },
+                "view_style_channel": {
+                    "type": "text"
+                },
+                "view_style_downloads": {
+                    "type": "text"
+                },
+                "view_style_playlist": {
+                    "type": "text"
+                },
+                "grid_items": {
+                    "type": "integer"
+                },
+                "hide_watched": {
+                    "type": "boolean"
+                },
+                "show_ignored_only": {
+                    "type": "boolean"
+                },
+                "show_subed_only": {
+                    "type": "boolean"
+                },
+                "sponsorblock_id": {
+                    "type": "text"
+                }
+            },
+            "expected_set": {
+                "analysis": {
+                    "normalizer": {
+                        "to_lower": {
+                            "type": "custom",
+                            "filter": [
+                                "lowercase"
+                            ]
+                        }
+                    }
+                },
+                "number_of_replicas": "0"
+            }
+        },
+        {
             "index_name": "channel",
             "expected_map": {
                 "channel_id": {
@@ -62,7 +120,7 @@
                             "type": "boolean"
                         },
                         "integrate_sponsorblock": {
-                            "type" : "boolean"
+                            "type": "boolean"
                         }
                     }
                 }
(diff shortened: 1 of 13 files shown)


bbilly1:
Sorry, only had a quick look. But at first glance, that goes into the right direction. :-)

A few thoughts:
- Migrations like that have been painful in the past, because it's unclear if you have already performed the migration or not, depending when that is getting called. That's why I have been doing this from a manage.py command, as [here](https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/config/management/commands/ta_startup.py#L150) for example or [here](https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/config/management/commands/ta_migpath.py). That then gets called in the startup [script](https://github.com/tubearchivist/tubearchivist/blob/master/docker_assets/run.sh). Doing this at startup is easier to check in the logs, you can have a nice message there. Plus you can be sure everybody running this will have the migration applied at the beginning and you can avoid complicated checking within the application itself. plus easy to remove again after a few versions, because all these things are nicely in one place.
- I see you found the index_mapping.json and how that gets applied. :-) It might be worth considering instead of having the complicated mapping where you define each config value as a field to just have a single text field with encoded JSON there. So basically the name would be `ta_config/_doc/user-1` containing a single text field with the encoded object for that user. then you can also have `ta_config/_doc/config` containing a single text field for the rest of the application config from downloads and stuff like that. So one field, one easy mapping, easy to extend in the future if we need to add an additional value as we won't have to rebuild the index. Only downside would be that you can update a single value, you'll have to get the whole config, decode the json, update the field, encode, and update it again. Does that make sense?

anonamouslyginger:
Understood, I will look into both of those things

anonamouslyginger:
Lots of manual testing aside, I think this is good to be reviewed.

There is other application configuration to move as well but this is stand-alone.

The only thing missing would be the comment on the migration of version from/to because I don't know what they are yet.

tubearchivist/home/src/es/index_mapping.json
```diff
@@ -1,5 +1,27 @@
 {
-    "index_config": [{
+    "index_config": [
```
bbilly1, line 2: Excellent choice on the object field type. I don't think you need the normalizer to lower there. That's needed in the other indexes for name keywords, but that doesn't apply here.

tubearchivist/home/src/es/index_mapping.json
```diff
@@ -62,7 +84,7 @@
                             "type": "boolean"
                         },
                         "integrate_sponsorblock": {
-                            "type" : "boolean"
+                            "type": "boolean"
```
bbilly1, line 87: please keep the json formatting as is. I know, people disagree but for me it looks strange to have a space before the `:`.

tubearchivist/home/src/ta/users.py
```diff
@@ -0,0 +1,136 @@
...
+            self._set_config("sponsorblock_id", sb_id)
+        return str(sb_id)
+
+    def get_view_style(self, view_name: str) -> str:
+        return str(self._get_config_or_default(f"view_style_{view_name}"))
+
+    def set_view_style(self, view_name: str, view_style: str):
+        self._set_config(f"view_style_{view_name}", view_style)
+
+    def _get_config_or_default(self, property: str) -> any:
```
bbilly1, line 102: `property` is a built in keyword in python. best to call it something else and not to overwrite it, to avoid any confusion later.

tubearchivist/home/src/ta/users.py
```diff
@@ -0,0 +1,136 @@
...
+        return self._config.get(property) or self._DEFAULT_USER_SETTINGS.get(
+            property
+        )
+
+    def _get_config(self) -> dict[str, any]:
```
bbilly1, line 107: nice, I've started adding typing for new code at least. For `any` you probably mean `from typing import Any` (note the uppercase)? as in the value of the dict can be of any data type?

bbilly1 (commented):
Very nice! I ran it, works flawlessly, at least from my testing. The object gets updated in ES as expected, one per user, very nice. :-)

We might also need a delete user conf command when you delete an user, just so it cleans up any leftover conf in ES too.

Added a few comments, but mostly cosmetics.

bbilly1:
Very nice! Thank you for staying on this. Merging now, can still do some testing for some edge cases going further before release.

PR #528, opened 2023-08-28
Split the settings page

anonamouslyginger:
As mentioned in #415 one of the outstanding tasks is to 'Split up settings page into separate sub pages'.

Here is my first attempt at that.
Assuming that I have captured everything I have created 4 views:
1. User Settings (the 'main' setting page at `/settings/`)
2. Application Settings (at `/settings/application/`)
3. Scheduling Settings (at `/settings/scheduling/`)
4. Actions (at `/settings/actions/`)

I have explicitly tried to not also solve the removal of POST forms to keep this PR well contained and 'simple'.
Please let me know if it matches your expectations @bbilly1.

See below the screenshots:
image
image
image
image

PR #535, opened 2023-09-03
Fix links to settings documentation

MerlinScheurer:
I noticed that the `Wiki` links where broken after the split by https://github.com/tubearchivist/tubearchivist/pull/528 & https://github.com/tubearchivist/docs/pull/19

And i made some of the links more precise.

---

For the url in `docs/Settings.md`  i was not sure which path to set,
because the `/settings/` is not a valid route anymore and returns `403 Forbidden`

### DIFF
--- a/tubearchivist/home/templates/home/settings_actions.html
+++ b/tubearchivist/home/templates/home/settings_actions.html
@@ -12,7 +12,7 @@
 </div>
 <div class="settings-group">
     <h2>Manual media files import.</h2>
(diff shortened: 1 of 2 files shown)


bbilly1:
good catch, forgot to update them. I'd say it's time to retire these old wiki pages, I think that's now 6 months plus since we moved things to our docs platform. So that would be as good a time as any to get rid of that. If some old links still point to the wiki, the docs will be right there on top of the readme. Do you want to go ahead and remove the docs folder from the repo there? Then I'll also deactivate the wiki.

But obviously great looking out on fixing the links in the html there.

bbilly1:
ah, just need to remove your changes on docs/settings from here to be able to merge that. I've forwarded testing branch.

bbilly1:
yeah, you are right, that would have been better. In any case, thanks for catching it.

PR #536, opened 2023-09-03
Refac biggest channels to be 3 seperate tables with ordering and right align

MerlinScheurer:
Right align numbers to make them a bit more comparable.

---

| Before | After |
| ------------- | ------------- |
| dashboard_(iPhone SE)  | dashboard_(iPhone SE) |
| dashboard_(Pixel 5)  | dashboard_(Pixel 5)  |
| dashboard_(iPad Air)  | dashboard_(iPad Air) |
| dashboard_settings_(full sized) | dashboard_(full size) |

Screenshot device order:
- iPhone SE
- Pixel 5
- iPad Air
- 2560 x 1440

Edit: Updated "after" screenshots

### DIFF
--- a/tubearchivist/api/src/aggs.py
+++ b/tubearchivist/api/src/aggs.py
@@ -196,6 +196,9 @@ class DownloadHist(AggBase):
 class BiggestChannel(AggBase):
     """get channel aggregations"""
 
+    def __init__(self, order):
+        self.data["aggs"][self.name]["multi_terms"]["order"] = { order: "desc" }
+
     name = "channel_stats"
     path = "ta_video/_search"
     data = {
--- a/tubearchivist/api/src/aggs.py
+++ b/tubearchivist/api/src/aggs.py
@@ -197,7 +197,7 @@ class BiggestChannel(AggBase):
     """get channel aggregations"""
 
     def __init__(self, order):
-        self.data["aggs"][self.name]["multi_terms"]["order"] = { order: "desc" }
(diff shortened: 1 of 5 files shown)


bbilly1:
I'll take a closer look at this later, but makes sense from your screenshots there, thanks for improving on that. hmm table is not ideal on mobile... should we just overflow this and make it scroll? Or something more smart like flexboxes where we can adjust the flow on mobile?

MerlinScheurer:
So, i reworked it a "little" bit 😅 

With help of a friend i got the order in the endpoint working
and splitted the table into 3 separate rankings ordered by their value.

Very long channel names might still cause problems, but there is more room for them now 😄

bbilly1:
Hmm, so to clarify, the biggest channels at the moment is the channel with the most videos. Then from that channel you also get duration and file size. But this will not be the channel with the biggest filesize for example. to actually get the biggest channels by filesize, that will need a different query.

MerlinScheurer:
Its doing 3 requests + elasticsearch-queries, to get the different rankings.

In the new screenshots you can see that the channels are in different order, because the order is based on the aggregation data returned from elasticsearch, which takes the `order_choices = ["doc_count", "duration", "media_size"]` into account.

MerlinScheurer:
Fixed the linting and rebased on the testing branch 😃

bbilly1:
excellent! Please ping me on Discord, as you see, there are many opportunities to shine in this project. :-)

PR #539, opened 2023-09-04
View only user

stratus-ss:
This PR makes some small changes to the `base.html` `video.html` and `channel.html`.

The purpose of this is to allow for essentially a view-only user so that videos cannot be deleted, channels cannot be added or other destructive actions.

This is accomplished by simply hiding the UI elements that get displayed depending on the following conditions

A) the user is apart of the "staff" built in Django option in the `/admin` page
B) is part of a group called "admin"

The `auth_extras` custom template tag was added in order to facilitate this. 

This PR is part of several smaller PRs that aim to make Tubearchivist a multi-user capable project

### Testing Criteria. 

1. Setup users similar to the following

image

2. Create the admin group:

image

3. Using private browsing for testing is recommended, log in as the admin user, you should see the download menu item:

image

The videos themselves should display index, delete and download options:

image

4. Login as a normal user, the download menu is gone:

image

The Channels menu should see the "+" sign removed indicating you cannot add a channel

image

NOTE: for this I am not using any special permissions in Django, it is enough that the group "admin" or the "is staff" are present in order to unlock the proper options

TODO: In another, smaller PR, add an indication as to which user you are logged in as currently


bbilly1:
Thanks, that's great progress! If I may ask, what's the case for the admin group? With the default django functionality you have:

- Superuser: Can do everything, including creating/modifying users
- Staff: Is a read/write user from the TA point of view, as in can create and modify channels etc...
- none of the above: Is a read only user.

From what I can see, this should cover all use cases? Or am I missing something?

And then this will also have to be verified server side too. I don't think we strive for a high security application, but functionality like that should also be handled in the API as well, while the markup changes is the UI/UX part of that.

stratus-ss:
> Thanks, that's great progress! If I may ask, what's the case for the admin group? 
> 
> 
> From what I can see, this should cover all use cases? Or am I missing something?

The thought was due to the SSO (LDAP) functionality you dont really want to manage users in LDAP and then in the application if you can. Even still if you are using LDAP you are more likely to just want to organize in groups as opposed to editing each user.

Regarding the API, can you point me at where I could find this code. I am sure I could dig it up, but if you happen to know where this might be implemented I can get started on that

bbilly1:
ah, yeah, SSO you are right, thanks for reminding me. :-)

API views are in a separate django app: https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/api/views.py, to have them separate from the regular views. 

You could create a mixin that checks for permissions and raises a 403 Forbidden? Or implement this in the Baseview and a class attribute? Instinctive I'd suggest the second approach but both could work.

stratus-ss:
I started poking at this and what I came up with was to create a function outside the class definition in `views.py`. The reason is that way all of the classes can make use of it. 

Currently looks like this:

```

def is_user_authorized(request, group_name):
        """Returns a bool after checking if request.user is either part of 
        the specified group OR has the is_staff django builtin

        Args:
            group_name (string): The name of the group to check for membership of

        Returns:
            bool: Returns true if user is part of group or is_staff django builtin
        """
        if request.user.groups.filter(name=group_name).exists() or request.user.is_staff:
            return True
        else:
            return False
```

It's possible this could be turned into a decorator if desired. I was having some problems with a decorator that was taking 'admin' as an argument due to the fact that `views.py` is being imported by `urls.py`. When this happens the class and method creating the decorator are not instantiated before they are referenced by the import. This was causing a "missing 1 positional argument error". I suppose it's possible to have the decorator hardcoded to just check for an `admin` group and thus would not need an argument.

I am currently applying the function at the top of certain methods like this:

```
    def delete(self, request, channel_id):
        # pylint: disable=unused-argument
        """delete channel"""
        if not is_user_authorized(request, 'admin'):
            raise PermissionDenied()
````

I decided against using a mixin for a few reasons. Thoughts on this approach?

**Comment by bbilly1** [comment]
I've been doing some research. I think decorators would be a more strait forward approach and will avoid code duplication and help with future maintainability. Django has this nifty shortcut decorator [user_passes_test](https://docs.djangoproject.com/en/4.2/topics/auth/default/#django.contrib.auth.decorators.user_passes_test) where you can pass it a function that takes the user object as a first argument and should return a boolean depending if the user passes. E.g. from your example above that function could look like:

```python
def check_admin(user):
    """if user has access to restricted views"""
    return user.is_staff or user.groups.filter(name="admin").exists()
```

Then to use decorators on class based views you'll have to wrap them in [method_decorator](https://docs.djangoproject.com/en/4.2/topics/class-based-views/intro/#decorating-the-class) then you can decorate the class itself to restrict all methods of that class, or a specific method alone, e.g. restrict only POST requests.

So your decorator could look something like this:

```python
@method_decorator(user_passes_test(check_admin), name='dispatch')
```

This would make things easier to extend, maybe at some point in the future we'll have to tweak the `check_admin` function there to pass additional groups, so we'll only have to do this in one place. Or maybe we'll want to add an additional level of restriction, so we can just easily create another decorator. 

Additionally same would also apply to the API views, so you can reuse that functionality. AFAIK, API requests will return 403, and regular view requests will redirect to the login page where you could login with a user that has access.

stratus-ss:
In the interest of being able to extend things in the future, I wonder if the right approach isnt putting this in a wrapper function like so:

```
def admin_only(function):
    @login_required
    @user_passes_test(lambda user: user.is_staff or user.groups.filter(name='admin').exists())
    def wrapper(*args, **kwargs):
        return function(*args, **kwargs)
    return wrapper
 ```

Particularly because that moves the `user_passes_test` into the function itself while at the same time allowing for more easy code refactor in the future...

Thoughts?

**Comment by bbilly1** [comment]
I think that would be overkill, at least for what I can foresee how this could be used and extended in the future. I'd say stick to the `check_admin` function as above and we should be good, even if we need to extend this in the future.

**Comment by stratus-ss** [comment]
this is being held up by https://github.com/tubearchivist/tubearchivist/issues/564

**Comment by stratus-ss** [comment]
@bbilly1 maybe we should merge this and I can work on the unittests anyways that way this gets in before there are more MCs. Thoughts?

**Comment by bbilly1** [comment]
Sounds good, I'll wont be able to review things before the weekend though...

**Comment by stratus-ss** [comment]
@bbilly1 Can you explain the linting failure on the views.py? I don't understand what I am looking at here. The other linting is obvious there needs to be a newline inserted

**Comment by lamusmaser** [comment]
The other one wants double-quotes instead of single-quotes.

**Comment by stratus-ss** [comment]
thanks i missed that!

**Comment by bbilly1** [comment]
you can run the linter locally too: https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#code-formatting-and-linting

Unfortunately I need to trigger it manually here.

**Comment by bbilly1** [comment]
Please run:

```bash
deploy.sh validate
```

Until that doesn't throw any errors any more. 

This executes this: https://github.com/tubearchivist/tubearchivist/blob/master/deploy.sh#L79-L102. So you can lint everything correctly.

stratus-ss:
Sorry about that. Too many things going on this week and I didn't devote the attention this needed

```
stratus@stratus-laptop ~/git_projects/tubearchivist  $ ./deploy.sh validate
run validate on .
running black
All done! ✨ 🍰 ✨
74 files would be left unchanged.
running codespell
running flake8
0
running isort
Skipped 2 files
    
> all validations passed
```

bbilly1:
OK, I'd say let's merge this and I'll do some fine tuning.

Thanks a lot for your addition, I'm sure many people will appreciate this.

Issue #462, opened 2023-04-10
[Feature Request]: Option to skip register backup repo at boostrap


MichaelKirgus:
### I've read the documentation

- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [feature request](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#feature-request) section.

### Your Feature Request

In my first attempt to set up this appliance, I tried to use my existing Elasticsearch cluster (13 nodes).
This fails because on startup the container wants to register a new repo to back up the indexes.
Unfortunately, this leads to the container not being started.
However, a powerful backup plan or strategy is usually already configured for such a multi-node instance.

The backup of the index can be switched off in the settings. Why not build in a way to skip this right at the first launch?
The user is then of course responsible for backing up his data.

I would like such a function for the container bootstrap process.
An idea here would be an optional environment variable, such as "SKIP_REGISTER_ES_REPO", with which this process is skipped.
Alternatively, simply do not let the start process abort, but issue a warning and then refer to the unsuccessfully registered backup repository in the settings.

### Your help is needed!

- [ ] Yes I will work on this in the next few days or weeks.

bbilly1:
v0.4.2 brings additional configuration options to set where to store the snapshots with a `ES_SNAPSHOT_DIR` env var. This should cover custom setups like that.

PR #551, opened 2023-09-15
Support elasticsearch clusters

ajgon:
This project is very opinionated when it comes for handling elasticsearch, and assumes only one ES node on one host. However it will be nice, to also support elasticsearch clusters (for example in k8s environments). This PR adds small tweaks here and there to make it possible.

**This PR is 100% backwards compatible.** All new settings default to previous logic.

1. Better snapshot checker

    The part of code checking for snapshots, doesn't take `role` param of ES node into consideration. It expects, all ES nodes to support snapshots, which is not the case in clustered environment. Only `data_*` and `master` nodes can have snapshots, other ones (especially `coordinating` and `ingest` nodes) doesn't allow it, so for these this check will fail. The PR takes that into consideration, and do the checks only on master/data nodes.

2. Customizable snapshot dir

    Snapshot dir is currently hardcoded to `/usr/share/elasticsearch/data/snapshot`. This is not always the case, and depends heavily on ES cluster configuration. I introduced new env variable `ES_SNAPSHOT_DIR` which defaults to `/usr/share/elasticsearch/data/snapshot` (for backwards compatibility) but also allows to customize this path.

3. Support for unverified SSL in elasticsearch

    Currently SSL is allowed for ES endpoint (by simply providing `https:` scheme), but requires trusted certificate - which may not always be the case (especially in internally-networked environments like kubernetes). The new `ES_VERIFY_SSL` allows to disable this verification while still using SSL. It must be explicitly set to `false` - everything else (including not setting it at all) will keep it enabled (as it was before).

### DIFF
--- a/tubearchivist/home/src/es/connect.py
+++ b/tubearchivist/home/src/es/connect.py
@@ -27,7 +27,7 @@ class ElasticWrap:
         self.url: str = f"{self.ES_URL}/{path}"
         self.auth: tuple[str, str] = (self.ES_USER, self.ES_PASS)
 
-        if not self.ES_DISABLE_VERIFY_SSL:
+        if self.ES_DISABLE_VERIFY_SSL:
             urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
     def get(
--- a/tubearchivist/home/src/es/connect.py
+++ b/tubearchivist/home/src/es/connect.py
@@ -9,6 +9,7 @@ import json
 import os
 
 import requests
+import urllib3
 
 
 class ElasticWrap:
@@ -19,19 +20,33 @@ class ElasticWrap:
     ES_URL: str = str(os.environ.get("ES_URL"))
     ES_PASS: str = str(os.environ.get("ELASTIC_PASSWORD"))
     ES_USER: str = str(os.environ.get("ELASTIC_USER") or "elastic")
+    ES_VERIFY_SSL: str = str(os.environ.get("ES_VERIFY_SSL") or "true")
 
     def __init__(self, path):
         self.url = f"{self.ES_URL}/{path}"
         self.auth = (self.ES_USER, self.ES_PASS)
+        self.verify_ssl = self.ES_VERIFY_SSL != "false"
+
+        if not self.verify_ssl:
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
     def get(self, data=False, timeout=10, print_error=True):
         """get data from es"""
         if data:
             response = requests.get(
-                self.url, json=data, auth=self.auth, timeout=timeout
+                self.url,
+                json=data,
+                auth=self.auth,
+                timeout=timeout,
+                verify=self.verify_ssl,
             )
         else:
-            response = requests.get(self.url, auth=self.auth, timeout=timeout)
+            response = requests.get(
+                self.url,
+                auth=self.auth,
+                timeout=timeout,
+                verify=self.verify_ssl,
+            )
         if print_error and not response.ok:
             print(response.text)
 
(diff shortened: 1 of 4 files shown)


bbilly1:
Thanks for looking into this. Please rebase on testing, there were some incoming changes there.

So if I'm understanding this correctly, by specifying `ES_SNAPSHOT_DIR` this would still allow the snapshot functionality to work identically as in a single node setup? As the problem described in #462.

bbilly1:
I think the desire to disable snapshots comes from the problem that this path wasn't configurable before for a cluster setup or any other custom installation.

What I meant with my comment was, this would work identically in a cluster setup as in a none cluster setup? All functionality works, as in autosnapshot on index change? create a snapshot on demand from the settings page as well as the daily policy trigger in the background?

At some point this deserves some nicer kwargs building there in the `ElasticWrap` class, but no need to do this now. :-)

ajgon:
I believe so. I'm not an ElasticSearch expert, but enabling snapshots works the same way in clustered setup. So as long as tubearchivist require snapshots, it is possible to enable them on any ES configuration. Configurable snapshot path only adds flexibility, for ES admin to choose where they store the snapshots.

bbilly1:
OK, please hear me out on this one, let's take the opportunity to refactor that `ElasticWrap` class there to dynamically build these keyword arguments. This would avoid the double negative there for the `ES_VERIFY_SSL` and we can be more descriptive what that does as this should disable SSL verification, we could set it directly like this:
```python
ES_DISABLE_VERIFY_SSL: bool = bool(os.environ.get("ES_DISABLE_VERIFY_SSL"))
```

So if you set that env var, this will result to `True`, which disables the verification, while keeping the current default behavior. 

then build the kwargs dynamically, e.g. in the `.get` method this would look something like:
```python
def get(self, data=False, timeout=10, print_error=True):
    """get data from es"""
    kwargs = {
        "auth": self.auth,
        "timeout": timeout,
    }

    if self.ES_DISABLE_VERIFY_SSL:
        kwargs["verify"] = False

    if data:
        kwargs["json"] = data

    response = requests.get(self.url, **kwargs)

    if print_error and not response.ok:
        print(response.text)

    return response.json(), response.status_code
```

then similar to that in the additional methods of that class too.

Are you up for that? If not, I can also push that to your PR, but I obviously don't want to take away from your great accomplishments with this addition here.

ajgon:
I don't feel very confident with python, so if you can - go for this refactor. It will be definitely better than whatever I will come up with :)

bbilly1:
no problem, just pushed the changes. please give it another try to verify.

tubearchivist/home/src/es/connect.py
```diff
@@ -19,65 +21,119 @@ class ElasticWrap:
...
-    def __init__(self, path):
-        self.url = f"{self.ES_URL}/{path}"
-        self.auth = (self.ES_USER, self.ES_PASS)
+    def __init__(self, path: str):
+        self.url: str = f"{self.ES_URL}/{path}"
+        self.auth: tuple[str, str] = (self.ES_USER, self.ES_PASS)
 
-    def get(self, data=False, timeout=10, print_error=True):
+        if not self.ES_DISABLE_VERIFY_SSL:
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
ajgon, line 31:
```suggestion
        if self.ES_DISABLE_VERIFY_SSL:
            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```

ajgon:
Deployed this PR and the forward-auth one to my cluster, and I can confirm everything works as expected 👍 . I just fixed minor issue [here](https://github.com/tubearchivist/tubearchivist/pull/551/commits/6abec9401bfaf4ee4b773cb1e41aa943d5f9a59e) (urllibg3 warning should be muted with SSL verification disabled, not enabled - meaning `ES_DISABLE_VERIFY_SSL` has to be set to `True` to mute it).

bbilly1:
Nice, thanks a lot for your contribution!

Issue #403, opened 2023-01-10
[Feature Request]: Remote user support


thijsa:
### Already implemented?

- [X] I have read through the [wiki](https://github.com/tubearchivist/tubearchivist/wiki).
- [X] I understand the [scope](https://github.com/tubearchivist/tubearchivist/wiki/FAQ) of this project and am aware of the [known limitations](https://github.com/tubearchivist/tubearchivist#known-limitations) and my idea is not already on the [roadmap](https://github.com/tubearchivist/tubearchivist#roadmap).

## Is your feature request related to a problem? Please describe.
I am using authentik (https://goauthentik.io/) in front of Tube Archivist. This handles the authentication for the users and is capable of sending some authentication info using headers to the application. Currently, users always need to sign in using the credentials in Tube Archivist after authenticating in authentik. 

## Describe the solution you'd like
Django has support for using the authentication information sent by the reverse proxy. See https://docs.djangoproject.com/en/4.1/howto/auth-remote-user/. I would like a setting / env variable that would allow enabling this feature so that users don't have to enter their credentials twice.

## Additional context
This feature should not be enabled by default because it would allow anyone to login with just the username if they don't have a proper reverse proxy setup. This warning is also present on the Django documentation page.

### Your help is needed!

- [x] Yes I can help with this feature request!

ionlights:
@thijsa Have you looked into using [`django-allauth`][allauth]? (This is slightly tangential to `REMOTE_USER` – if `tubearchivist` used `django-allauth` this would allow for OIDC authN/authZ.)

[Tandoor (Recipes)][tandoor] (which is also a Django app) uses `django-allauth`. A recent PR to `django-allauth` allowed for generic OIDC (which means Authentik now have proper support), I've been using it for a few weeks without issue.

[allauth]: https://django-allauth.readthedocs.io/en/latest/
[tandoor]: https://github.com/tandoorrecipes/recipes/

zerosnugget:
> @thijsa Have you looked into using [`django-allauth`](https://django-allauth.readthedocs.io/en/latest/)? (This is slightly tangential to `REMOTE_USER` – if `tubearchivist` used `django-allauth` this would allow for OIDC authN/authZ.)
> 
[...]

PR #552, opened 2023-09-15
Support for forward auth

ajgon:
This PR adds support for authentication proxies such as Authelia.

### DIFF
--- a/tubearchivist/home/urls.py
+++ b/tubearchivist/home/urls.py
@@ -3,18 +3,30 @@
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.views import LogoutView
+from django.shortcuts import redirect
 from django.urls import path
 from home import views
 
-urlpatterns = [
-    path("", login_required(views.HomeView.as_view()), name="home"),
-    path("login/", views.LoginView.as_view(), name="login"),
-    path(
+if hasattr(settings, "TA_AUTH_PROXY_LOGOUT_URL"):
+    logout_path = path(
+        "logout/",
+        lambda request: redirect(
+            settings.TA_AUTH_PROXY_LOGOUT_URL, permanent=False
+        ),
+        name="logout",
+    )
+else:
+    logout_path = path(
         "logout/",
         LogoutView.as_view(),
         {"next_page": settings.LOGOUT_REDIRECT_URL},
         name="logout",
-    ),
+    )
+
+urlpatterns = [
+    path("", login_required(views.HomeView.as_view()), name="home"),
+    path("login/", views.LoginView.as_view(), name="login"),
+    logout_path,
     path("about/", views.AboutView.as_view(), name="about"),
     path(
         "downloads/",
--- a/README.md
+++ b/README.md
@@ -58,6 +58,9 @@ Take a look at the example [docker-compose.yml](https://github.com/tubearchivist
 | TZ | Set your timezone for the scheduler | Required |
 | TA_PORT | Overwrite Nginx port | Optional |
 | TA_UWSGI_PORT | Overwrite container internal uwsgi port | Optional |
+| TA_ENABLE_AUTH_PROXY | Enables support for forwarding auth in reverse proxies | [Read more](https://docs.tubearchivist.com/configuration/forward-auth/) |
+| TA_AUTH_PROXY_USERNAME_HEADER | Header containing username to log in | Optional |
+| TA_AUTH_PROXY_LOGOUT_URL | Logout URL for forwarded auth | Opttional |
 | ES_URL | URL That ElasticSearch runs on | Optional |
 | ES_DISABLE_VERIFY_SSL | Disable ElasticSearch SSL certificate verification | Optional |
 | ES_SNAPSHOT_DIR | Custom path where elastic search stores snapshots for master/data nodes | Optional |
--- a/tubearchivist/config/settings.py
+++ b/tubearchivist/config/settings.py
@@ -175,7 +175,6 @@ if bool(environ.get("TA_LDAP")):
             ldap.OPT_X_TLS_REQUIRE_CERT: ldap.OPT_X_TLS_NEVER,
         }
 
-    global AUTHENTICATION_BACKENDS
     AUTHENTICATION_BACKENDS = ("django_auth_ldap.backend.LDAPBackend",)
 
 # Database
@@ -210,6 +209,19 @@ AUTH_PASSWORD_VALIDATORS = [
 ]
 
 AUTH_USER_MODEL = "home.Account"
+
+# Forward-auth authentication
+if bool(environ.get("TA_ENABLE_AUTH_PROXY")):
+    TA_AUTH_PROXY_USERNAME_HEADER = (
+        environ.get("TA_AUTH_PROXY_USERNAME_HEADER") or "HTTP_REMOTE_USER"
+    )
+    TA_AUTH_PROXY_LOGOUT_URL = environ.get("TA_AUTH_PROXY_LOGOUT_URL")
+
+    MIDDLEWARE.append("home.src.ta.auth.HttpRemoteUserMiddleware")
+
+    AUTHENTICATION_BACKENDS = (
+        "django.contrib.auth.backends.RemoteUserBackend",
+    )
 
 
 # Internationalization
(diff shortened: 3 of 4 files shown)


bbilly1:
Thanks for taking the initiative on this.

Does this cover #403?

@DanielBatteryStapler you might want to take a quick look at this to verify LDAP compatibility.

Could both backends be used in parallel? As some users authenticate with LDAP and some with REMOTE_USER?

In anycase, please also make a PR against the docs.

DanielBatteryStapler:
Doesn't seem to cause any LDAP problems that I can tell.

Django does officially support having multiple authentication backends enabled at the same time, how it works is that when a user attempts a login it checks to see if it can authenticate using any of the backends, so I don't see why having both enabled at the same time shouldn't work -- by work I mean not immediately crash. That said, an authenticating proxy like Authelia is going to pretty much completely preempt LDAP, if I'm understanding correctly, as once you authenticate with Authelia you'll be considered logged in by tubearchivist and there would be no way to try and login using your LDAP account anyways. I don't think this is an issue though. If you're an admin setting up this to be used with something like Authelia I don't know why you would want some users to be able to circumvent that and use LDAP instead. I can't think of any reasonable way for that to work. Similarly the way I implemented LDAP support also disables the default local authentication that tubearchivist uses as otherwise it would check users against both the LDAP and Django's local database instead of just using LDAP, so I think it also makes sense for this proxy authentication to do the same thing.

In summary, I think it should be an either/or situation with the LDAP and authentication proxy similar to how it is an either/or with LDAP and local authentication. Having two or more enabled doesn't make sense to me even if Django technically allows it.

All of that said, I think that this PR actually breaks local authentication because [it is setting AUTHENTICATION_BACKENDS](https://github.com/tubearchivist/tubearchivist/pull/552/files#diff-bd0c50d1b0797afb8f8c8f55b987e1fe6d252ca7745d9f325634c46931439cf9R88) to empty array when neither LDAP or proxy authentication is enabled when Django's default is actually [["django.contrib.auth.backends.ModelBackend"]](https://docs.djangoproject.com/en/4.2/topics/auth/customizing/). I'm pretty sure it's necessary for the array to be set to that in order for local authentication to work. I also think that the .insert(0,...)s are unnecessary as it doesn't make sense for more than one authentication backend (as far as I know, if I'm wrong please tell me) and you should just completely set the array to the one backend you actually want to use. If both LDAP and the authentication proxy are enabled that would mean that authentication proxy takes precedence, but you can just put a check in the code for that and print an error or something, I don't think it's really that important. The admin should know what they're doing, but I'd prefer the code not suggest that having more than one enabled actually makes sense.

All of that said, I haven't actually tested the code on my computer, this is just what I noticed when I read through it. Also, just to make sure I'm not coming off the wrong way, this isn't my repo, this is bbilly1's, so don't take my suggestions as gospel, but that's my thoughts on these changes. Besides my suggestions, I think everything looks great and I'm glad to see Authelia get support.

ajgon:
@DanielBatteryStapler this makes a lot sense, thanks! I'm not that familiar with Django authentication (to be honest, I'm not even Python dev :-) ). But I agree, emptying array is a bug - I'll remove that, and follow your advice, to override it on ForwardAuth part. 

I agree, that it doesn't make sense to enable local authentication when forrwad auth is enabled, as user has to authenticate via proxy (authelia for example) in a first place. Asking him to authenticate again (this time locally), doesn't make sense in my opinion.

DanielBatteryStapler:
The changes you just made probably don't work because you're missing the global line that was there before (needed because Python). I think you mean to leave the ldap one as
`global AUTHENTICATION_BACKENDS
 AUTHENTICATION_BACKENDS = ("django_auth_ldap.backend.LDAPBackend",)`
 and have the other one be
 `global AUTHENTICATION_BACKENDS
 AUTHENTICATION_BACKENDS = ("django_auth_ldap.backend.RemoteUserBackend",)`
 
 Unless I'm wrong and it works without the `global` declaration, but I'm pretty sure that's required.

ajgon:
I remember when I was testing this, that two globals for the same option was failing (despite being in the exclusive IF statements). Let me test it, and fix it (maybe extracting global outside be sufficient).

DanielBatteryStapler:
Yeah the `global AUTHENTICATION_BACKENDS` lines clarify to Python that you aren't trying to make a new local variable, you are trying to set the variable globally

tubearchivist/home/urls.py
```diff
@@ -3,18 +3,30 @@
...
 from django.contrib.auth.views import LogoutView
+from django.shortcuts import redirect
 from django.urls import path
 from home import views
 
-urlpatterns = [
-    path("", login_required(views.HomeView.as_view()), name="home"),
-    path("login/", views.LoginView.as_view(), name="login"),
-    path(
+if settings.TA_AUTH_PROXY_LOGOUT_URL:
```
bbilly1, line 10: This will raise an `AttributeError` if you don't configure that. You'll have to check that attribute first before you can access it, e.g. something like this should work:
```python
hasattr(settings, "TA_AUTH_PROXY_LOGOUT_URL")
```
That will return `True` if the attribute is set in `settings.py`, `False` otherwise.

bbilly1 (changes requested):

bbilly1:
@ajgon Small fix needed, you'll need to check if an attribute is set, before you can access it. 

Thanks @DanielBatteryStapler for your insights. If you have the time, please give it a proper test to verify all is working on your end as expected. 

Alternatively I'll also make an unstable build after merging, you could test it then too, so we can catch any problems before this goes out as a release.

ajgon:
@bbilly1 Added `hasattr` patch, rebased everything back to `testing` and checked on my env new change (it works) :-)

DanielBatteryStapler:
I tested the LDAP support against my own server using this docker-compose (basically unchanged since LDAP support was originally added)
```
version: '3.3'

services:
  
  daniel-authenticator:
    restart: unless-stopped
    tty: true
    stdin_open: true
    container_name: daniel-authenticator
    image: danielbatterystapler/daniel-authenticator
    environment:
      DANIEL_AUTHENTICATOR_SECRET_KEY: "thishouldbeaverysecretrandomseriesofcharacters"
      DANIEL_AUTHENTICATOR_DECLARATIVE_DATABASE: |
        create_user john "John Admin" johnsmith@example.com secret_password_john
        set_user_uuid john 75e50c72-2801-4c1d-ab3c-2f77fed25d10
        superuser john
        
        create_user mark "Mark User" markuser@example.com secret_password_mark
        set_user_uuid mark 891801a2-d162-4900-96ea-77c15c30a864
        
        create_service tubearchivist_service "TubeArchivist" "http://localhost:8000" secret_password_service
        add_user_to_service john tubearchivist_service
        add_user_to_service mark tubearchivist_service
        
    ports:
      - 8080:8080

  tubearchivist:
    container_name: tubearchivist
    restart: unless-stopped
    build:
      context: ./tubearchivist/
    ports:
      - 8000:8000
    environment:
      ES_URL: http://archivist-es:9200  
      REDIS_HOST: archivist-redis       
      HOST_UID: 1000
      HOST_GID: 1000
      TA_USERNAME: john       
      TA_PASSWORD: verysecret   # notice that this password doesn't actually let you login, you need to use the one specified in the ldap service
      TA_HOST: localhost
      ELASTIC_PASSWORD: verysecret     
      TZ: America/New_York      
      
      TA_LDAP: "true"
      TA_LDAP_SERVER_URI: "ldap://daniel-authenticator:3389"
      TA_LDAP_BIND_DN: "ou=tubearchivist_service,ou=services,dc=daniel-authenticator"
      TA_LDAP_BIND_PASSWORD: "secret_password_service"
      TA_LDAP_USER_BASE: "ou=users,ou=tubearchivist_service,ou=services,dc=daniel-authenticator"
      TA_LDAP_USER_FILTER: "(objectClass=user)"
    depends_on:
      - archivist-es
      - archivist-redis

  archivist-redis:
    image: redislabs/rejson                 # for arm64 use bbilly1/rejson
    container_name: archivist-redis
    restart: unless-stopped
    expose:
      - "6379"
    depends_on:
      - archivist-es

  archivist-es:
    image: bbilly1/tubearchivist-es         # only for amd64, or use official es 7.17.4
    container_name: archivist-es
    restart: unless-stopped
    environment:
      "xpack.security.enabled": "true"
      ELASTIC_PASSWORD: "verysecret"
      discovery.type: "single-node"
      ES_JAVA_OPTS: "-Xms1024m -Xmx1024m"
      "path.repo": "/usr/share/elasticsearch/data/snapshot"
    ulimits:
      memlock:
        soft: -1
        hard: -1
    expose:
      - "9200"
```
and it  still works as expected. Looks good to me!

bbilly1 (approved):

PR #553, opened 2023-09-15
Healthcheck endpoint

ajgon:
This PR adds simple `/health` endpoint which is usable in docker swarm or kubernetes environments, which use periodical healthchecks of the container.

### DIFF
--- a/tubearchivist/config/settings.py
+++ b/tubearchivist/config/settings.py
@@ -64,6 +64,7 @@ MIDDLEWARE = [
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "home.src.ta.health.HealthCheckMiddleware",
 ]
 
 ROOT_URLCONF = "config.urls"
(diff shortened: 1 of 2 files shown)


bbilly1:
hmm, we already have a [`/api/ping/`](https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/api/views.py#L628-L641) view for such occasions. Maybe that would do aswell?

lamusmaser:
Clusters tend to default to a `/healthcheck` endpoint that responds with a simple `OK`, unless @ajgon notes otherwise. Haven't had to build one for a long while.

ajgon:
I'm fine with any name for the endpoint. I can rename it to `/healthcheck` as well :)

@bbilly1 - the problem with `/api/ping/` is, that is hidden behind authorization - and returns 403. Usually healthcheck endpoints expect one of the `2xx` (preferably `200`) statuses.

lamusmaser:
I'm good with `/health` or `/healthcheck`.

Since this project has dependencies, we may also need to show their status, unless you aren't worried about their up status since we can assume it from TA being online.

bbilly1:
Nice, sounds good. Please also make a mention in the docs, maybe with a docker-compose example similar to:
```yml
    healthcheck:
      test: ["CMD", "curl", "-f", "http://localhost:8000/health"]
      interval: 2m
      timeout: 10s
      retries: 3
      start_period: 60s
```
Then this should be enough for other platform who have their own health check functionality or such to build their own.

PR #555, opened 2023-09-19
fix links to wiki

macabeus:
This PR fixes the wiki links that are currently broken.


bbilly1:
thanks for looking into this. these were already fixed with #535, release is pending.

take a look at [CONTRIBUTING.md](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md), testing branch is the most uptodate for things like that.

PR #567, opened 2023-10-09
Remove /media/ prefix from Download File URL in video.html

OmarLaham:
Hello all,
I removed the /media/ prefix from Download File URL in the video.html. I left the search processor as it's following @bbilly1 's advice.
This is my first contribution to open source! If other modifications are required or I made mistakes please notify me.


bbilly1:
very nice, thanks for fixing it.

PR #571, opened 2023-10-14
Move the startup application settings to a new class

anonamouslyginger:
This is part two of moving the application configuration to Elasticsearch.

The `AppConfig` class contained some properties which do not change during the applications execution:

- Media Directory - Always `/youtube`
- Application Root - Always `/app`
- Cache Directory - Always `/cache`
- UID to store files on the host system - Set from the `HOST_UID` environment variable at start up
- GID to store files on the host system - Set from the `HOST_GID` environment variable at start up
- Chromecast enabled/disabled - Set from the `ENABLE_CAST` environment variable at start up

I have moved these to a separate class `EnvironmentSettings` to differentiate them from the other application settings which do change during execution.

**Requests for input**:

1. There's a mixture of form in the current classes (should properties like the `cache_dir` be a constant or a field) and I'm open to a suggestion on making them consistent.
2. Should this MR also set up a migration to delete the redundant properties from Redis?

**Note**: I do not own any Chrome cast devices so I have only tested that the HTML element appears in the source.

### DIFF
--- a/tubearchivist/api/src/search_processor.py
+++ b/tubearchivist/api/src/search_processor.py
@@ -7,15 +7,14 @@ Functionality:
 import urllib.parse
 
 from home.src.download.thumbnails import ThumbManager
-from home.src.ta.config import AppConfig
 from home.src.ta.helper import date_praser, get_duration_str
+from home.src.ta.settings import EnvironmentSettings
 
 
 class SearchProcess:
     """process search results"""
 
-    CONFIG = AppConfig().config
-    CACHE_DIR = CONFIG["application"]["cache_dir"]
+    CACHE_DIR = EnvironmentSettings().get_cache_dir()
 
     def __init__(self, response):
         self.response = response
--- a/tubearchivist/api/src/search_processor.py
+++ b/tubearchivist/api/src/search_processor.py
@@ -14,7 +14,7 @@ from home.src.ta.settings import EnvironmentSettings
 class SearchProcess:
     """process search results"""
 
-    CACHE_DIR = EnvironmentSettings().get_cache_dir()
+    CACHE_DIR = EnvironmentSettings.CACHE_DIR
 
     def __init__(self, response):
         self.response = response
--- a/tubearchivist/api/views.py
+++ b/tubearchivist/api/views.py
@@ -18,6 +18,7 @@ from home.src.index.playlist import YoutubePlaylist
 from home.src.index.reindex import ReindexProgress
 from home.src.index.video import SponsorBlock, YoutubeVideo
 from home.src.ta.config import AppConfig, ReleaseVersion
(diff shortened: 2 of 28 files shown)


bbilly1:
You are on a roll! :-)

but man, does this tie into many things, hmm? :-)

original thinking behind having these folder paths as part of the config was for a future none docker installation method. For that, you might want to be able to configure that. But you could still easily run this outside of docker by changing these values in the config. So how about we add some environment variables there?

In any case, I like what you are doing here, separating of concerns with the different part of the configs, very reasonable. As you noted, these are all set at application start up, so we don't have to read these values again. So you could simplify that `EnvironmentSettings` to look something like this:

```python
class EnvironmentSettings:

    HOST_UID: int = int(os.environ.get("HOST_UID", False))
    HOST_GID: int = int(os.environ.get("HOST_GID", False))
    ENABLE_CAST: bool = bool(os.environ.get("ENABLE_CAST"))
    MEDIA_DIR: str = str(os.environ.get("TA_MEDIA_DIR", "/youtube"))
    APP_DIR: str = str(os.environ.get("TA_APP_DIR", "/app"))
    CACHE_DIR: str = str(os.environ.get("TA_CACHE_DIR", "/cache"))

```
Then to access a value from there, for example:
```python
EnvironmentSettings.APP_DIR
```

So we don't need to read the environment over and over, as you saw, in quite a lot of different places throughout the application.

Hope this makes sense?

bbilly1:
Additionally, lets take the opportunity to to bundle some more of these environment variables there, so we have kind of a self documenting place of where these are getting set, including defaults if there are any.

Maybe check where else os.environ is imported and used and go from there? Some obvious one I found were `ELASTIC_USER`, `ES_URL`, `ELASTIC_PASSWORD`. `TA_PORT`, `TA_UWSGI_PORT`, `TA_USERNAME`, `TA_PASSWORD`, `REDIS_HOST`, `REDIS_PORT`... maybe you find some more. They are now spread out through the application and it's hard to track all of them, and what can be configured where, what the defaults are...

In some cases, maybe it would even make sense to have a `ElasticEnviron` class, or similar, for these?

anonamouslyginger:
Implemented most of the suggestions. Just going through `ta_envcheck.py` which is a bit trickier because we want to output what the env is set to and differentiate that from the default, which the above pattern hides behind a field with a default.

I think having a single location makes the most sense. There's not enough for it to become overwhelming to warrant a separate `ElasticEnviron`, for example.

anonamouslyginger:
I think I've centralised everything related to the application and this is ready to go

bbilly1:
This looks very nice! I'll be able to look at it closer on the weekend.

bbilly1:
beautifully done. can be merged as is.

PR #577, opened 2023-10-27
Refac settings dashboard

MerlinScheurer:
image

image


bbilly1:
Looks good on the JS side.

For 0 padding, you can use f string directly, e.g.:
```python
number = 1
print(f"label: {number:02}")
```
will print
```
label: 01
```

PR #583, opened 2023-11-04
Better CSS support

extome9000:
This PR removes hardcoded CSS stylesheets and allows for more stylesheets to be added. CSS files are scanned in `/app/static/css` (`style.css` is removed since it's the root stylesheet) and can be picked from the settings where the color scheme option used to be.

I have included 2 example stylesheets, `midnight.css` and `matrix.css`, to show how different styles could look.

Feedback is appreciated.

### DIFF
--- a/tubearchivist/home/src/frontend/forms.py
+++ b/tubearchivist/home/src/frontend/forms.py
@@ -6,6 +6,7 @@ from django import forms
 from django.contrib.auth.forms import AuthenticationForm
 from django.forms.widgets import PasswordInput, TextInput
 from home.src.ta.helper import get_stylesheets
+from os.path import splitext
 
 
 class CustomAuthForm(AuthenticationForm):
@@ -31,7 +32,7 @@ class UserSettingsForm(forms.Form):
     """user configurations values"""
 
     stylesheets = get_stylesheets()
-    CHOICES = [(stylesheet, stylesheet) for stylesheet in stylesheets]
+    CHOICES = [(stylesheet, splitext(stylesheet)[0].title()) for stylesheet in stylesheets]
     CHOICES.insert(0, ("", "-- change color scheme --"))
 
     stylesheet = forms.ChoiceField(
--- a/tubearchivist/home/src/frontend/forms.py
+++ b/tubearchivist/home/src/frontend/forms.py
@@ -32,15 +32,16 @@ class CustomAuthForm(AuthenticationForm):
 class UserSettingsForm(forms.Form):
     """user configurations values"""
 
-    stylesheets = get_stylesheets()
-    CHOICES = [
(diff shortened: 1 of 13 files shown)


tubearchivist/home/src/frontend/forms.py
```diff
@@ -29,13 +32,14 @@ class CustomAuthForm(AuthenticationForm):
 class UserSettingsForm(forms.Form):
     """user configurations values"""
 
+    stylesheets = get_stylesheets()
     CHOICES = [
```
bbilly1, line 36: You can simplify that with something like:
```python
class UserSettingsForm(forms.Form):
    """user configurations values"""

    STYLESHEET_CHOICES = [("", "-- change stylesheet --")]
    STYLESHEET_CHOICES.extend(
        [
            (stylesheet, os.path.splitext(stylesheet)[0].title())
            for stylesheet in get_stylesheets()
        ]
    )

    stylesheet = forms.ChoiceField(
        widget=forms.Select, choices=STYLESHEET_CHOICES, required=False
    )
    page_size = forms.IntegerField(required=False)
```
This also is more explicit what is a field and what is an additional form attribute.

tubearchivist/home/src/ta/helper.py
```diff
@@ -203,3 +204,24 @@ def ta_host_parser(ta_host: str) -> tuple[list[str], list[str]]:
...
+    """Get all valid stylesheets from /static/css"""
+    app_root = EnvironmentSettings.APP_DIR
+    stylesheets = os.listdir(os.path.join(app_root, "static/css"))
+    stylesheets.remove("style.css")
+    stylesheets.sort()
+    stylesheets = list(filter(lambda x: x.endswith(".css"), stylesheets))
+    return stylesheets
+
+
+def check_stylesheet(stylesheet: str):
```
bbilly1, line 219: You already implement the `get_stylesheets` functionality above, you can use that, something like this:
```python
def check_stylesheet(stylesheet: str):
    """Check if a stylesheet exists. Return dark.css as a fallback"""
    if stylesheet in get_stylesheets():
        return stylesheet

    return "dark.css"
```

bbilly1 (changes requested):

bbilly1:
Very cool, works great on some channels:
2023-11-08_22-36-34-screenshot

Also midnight is quite nice too!
2023-11-08_22-37-35-screenshot

I added two small comments with suggestions.

Great addition! Do you also want to make a PR against the docs? Add some instructions? Ask people to add their own addition.

extome9000:
I'll add a page for CSS files on the docs repository for how to make and add stylesheets. I've also committed your suggested changes.

bbilly1 (approved):

bbilly1:
Very well done. Thanks a lot for your contribution. :-)

PR #594, opened 2023-11-13
Fix `textarea` type from `shell` to `Shell`.

lamusmaser:
Fix `textarea` render type, changing it from `shell` to `Shell`. Noticed this issue when setting up other project (Plex integration).

### DIFF
--- a/.github/ISSUE_TEMPLATE/INSTALLATION-HELP.yml
+++ b/.github/ISSUE_TEMPLATE/INSTALLATION-HELP.yml
@@ -38,6 +38,6 @@ body:
     attributes:
       label: Relevant log output
       description: Please copy and paste any relevant Docker logs. This will be automatically formatted into code, so no need for backticks.
-      render: shell
+      render: Shell
     validations:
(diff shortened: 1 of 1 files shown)


bbilly1:
Not sure what issue this is solving? According to the docs on creating issue templates, this should be lower case: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

Id didn't see any problems with that so far, or what am I missing?

lamusmaser:
You are correct, that is what is used in their example. However, their validation comes through a [`github-issue-forms.json`](https://json.schemastore.org/github-issue-forms.json), and that one uses the capitalized version on line 609. This was identified as  non-conforming by VSCode YAML Extension.

As a note, it hasn't really been an issue since GitHub seems to be automatically determining the render value correctly, but this could cause issues with others in the future. I noticed it when I was validating inputs for TubeArchivist-Plex and confirmed it was here, as well.

Issue #581, opened 2023-11-01
[Bug]: Manual video import fails due to old metadata info json from an old youtube-dl version


dot-mike:
### I've read the documentation

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

Docker

## Describe the bug

Importing video manually fails. I have the video and the metadata-file from way back time but TA won't import it.
It was downloaded with an old version (2020~) of youtube-dl

### Relevant log output

```shell
2023/11/02 00:44:08	stderr	KeyError: 'tags'
2023/11/02 00:44:08	stderr	            ~~~~~~~~~~~~~~~~~^^^^^^^^
2023/11/02 00:44:08	stderr	    "tags": self.youtube_meta["tags"],
2023/11/02 00:44:08	stderr	  File "/app/home/src/index/video.py", line 194, in _process_youtube_meta
2023/11/02 00:44:08	stderr	    self._process_youtube_meta()
2023/11/02 00:44:08	stderr	  File "/app/home/src/index/video.py", line 149, in build_json
2023/11/02 00:44:08	stderr	    video.build_json(
2023/11/02 00:44:08	stderr	  File "/app/home/src/index/manual.py", line 402, in index_metadata
2023/11/02 00:44:08	stderr	                ^^^^^^^^^^^^^^^^^^^^^
2023/11/02 00:44:08	stderr	    json_data = self.index_metadata()
2023/11/02 00:44:08	stderr	  File "/app/home/src/index/manual.py", line 394, in run
2023/11/02 00:44:08	stderr	    ManualImport(current_video, self.CONFIG).run()
2023/11/02 00:44:08	stderr	  File "/app/home/src/index/manual.py", line 146, in process_videos
2023/11/02 00:44:08	stderr	    self.process_videos()
2023/11/02 00:44:08	stderr	  File "/app/home/src/index/manual.py", line 55, in scan
2023/11/02 00:44:08	stderr	    ImportFolderScanner(task=self).scan()
2023/11/02 00:44:08	stderr	  File "/app/home/tasks.py", line 271, in run_manual_import
[...]

PR #595, opened 2023-11-13
Add `allowed_null_keys` and its dictionary for manual imports.

lamusmaser:
Create default dictionary for allowed null inputs within manual imports. Related to https://github.com/tubearchivist/tubearchivist/issues/581.

### DIFF
--- a/tubearchivist/home/src/index/manual.py
+++ b/tubearchivist/home/src/index/manual.py
@@ -427,17 +427,8 @@ class ManualImport:
         if not self.current_video["metadata"]:
             return False
 
-        allowed_null_keys = [
-            "description",
-            "categories",
-            "thumbnail",
-            "tags",
-            "view_count",
-        ]
-        default_null = dict.fromkeys(allowed_null_keys, None)
-
         with open(self.current_video["metadata"], "r", encoding="utf-8") as f:
-            info_json = json.loads(f.read(), object_hook=default_null)
+            info_json = json.loads(f.read())
 
         return info_json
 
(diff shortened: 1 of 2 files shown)


lamusmaser:
Sorry for the linter abuse, my local linter didn't complain about any of those issues.

bbilly1:
Thanks for looking into this. It would be better to implement that in `_process_youtube_meta` of the `YoutubeVideo` index class. So we can keep the expected datatype and not just set it to `None`. Current behavior of yt-dlp for example is, if there aren't any tags in a video, it will be an empty list `[]`. So it would be best to imitate the same behavior too. So this could look like this:
```python
self.json_data = {
    "title": self.youtube_meta["title"],
    "description": self.youtube_meta["description"],
    "category": self.youtube_meta.get("categories", []),
    "vid_thumb_url": self.youtube_meta["thumbnail"],
    "vid_thumb_base64": base64_blur,
    "tags": self.youtube_meta.get("tags", []),
    "published": published,
    "vid_last_refresh": last_refresh,
    "date_downloaded": last_refresh,
    "youtube_id": self.youtube_id,
    # Using .value to make json encodable
    "vid_type": self.video_type.value,
    "active": True,
}
```
Note the use of `.get()` with the fallback value. then same in `_add_stats` to use `0` for `view_count` fallback.

lamusmaser:
That's fair, I didn't want to change the base processor in `_process_youtube_meta` so that it could be untouched for YouTube interactions, but maybe that is a good fallback anyway. I can adjust the changes to be there instead of their current location.

bbilly1:
thanks for implementing that.

PR #598, opened 2023-11-15
Add new CSS item for footer text centering

derekslenk:
Add CSS style to center the text in the footer.

Text was left-aligned which on setups with more than 3 columns starts to feel out-of-place.

This matches the mobile style of centering the text as well.


derekslenk:
Pre fix, multi-colum multi-page is off center
image

Post fix, not enough videos for lots of pages
image

bbilly1:
nice, thanks for fixing that.

Issue #590, opened 2023-11-11
[Feature Request]: User created playlist


greg321321:
### I've read the documentation

- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [feature request](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#feature-request) section.

### Your Feature Request

I would like to build a Watch Offline feature into this project. 

Use Case: When at a location where there is no internet connection, I would like to view videos on my PED (Personal Electronic Device).

User Stories:

1. As a user, I want to create a custom playlist. (existing roadmap feature)
2. As a user, I want to toggle a custom playlist for Watch Offline.
3. As a user, when I am at a location where there is no internet connection, I want to view my 'Watch Offline custom playlist' videos.

I believe all this work can be done simply using web caching. Let me know if you think this is a good idea. I can build the feature with your direction and knock out one or two of your roadmap features/bugs in the process. I am a software developer familiar with web technologies/gitlab.

### Your help is needed!

- [X] Yes I will work on this in the next few days or weeks.

bbilly1:
Well, this *is* already an offline application. TA doesn't need internet connection to work, after you have downloaded and indexed your videos obviously. Should also be portable, like move a portion of TA to you laptop temporarily, all persistent data is in the volume, or if you know you'll be offline later, sync your data first, then you can use TA offline perfectly well. Just saying that you can do that already.

To the rest, obviously, please go ahead, any of the documented bug reports or feature requests are ready to be taken on by you.

greg321321:
ok I think I will start on the playlist idea first (existing roadmap feature).  The offline scenario is for an Android phone. I goto the gym and they don't have good WiFi, I don't have an unlimited cell data plan, but would still like to watch videos from my phone.

bbilly1:
Ah, I see, you mean offline from the TA server and not from the internet. Honestly that sounds like [out of scope](https://docs.tubearchivist.com/faq/#what-is-the-scope-of-this-project) for this project, I don't see how this could be done with caching alone, as the interface constantly interacts with the backend.

Maybe some sort of static site generator would be a better approach? Or some sort of export script that can just get the video files you want and makes a copy to your phone? So you can use the phones built in media player?

greg321321:
[...]

PR #602, opened 2023-11-16
Implement custom playlists

greg321321:
This is just an early preview draft of the functionality. You should be able to take the feature for a spin and provide me some feedback. Ticket #590 has screenshots of expected functionality. I still need to align the code to existing patterns in the project, disable debugging, remove unimplemented functions, update documentation, etc.


bbilly1:
It's worth looking into how I implemented the current playlist functionality and integrate as much of that as possible. 

If you look how I implemented that, you'll see the video document in ES has a playlist array, with a list of playlist IDs where this video can be found. Then the playlist document in ES has the playlist_entries, that only contains the minimum needed to basically build the playlist navigation and not the complete video metadata. This way you avoid duplication as much as possible and you can have separation of concern as much as possible.

So to implement custom playlists you don't need a separate index in ES for that, also you wouldn't want to duplicate the complete video metadata to playlist_entries, that is going to balloon your index as a video can be in multiple playlists, theoretically even the same video can be multiple times in the same playlist. Then you'll also have to consider other functionality of this application like reindex video metadata...

What you need is a way to differentiate between youtube and local playlist, you could do that by adding a bool flag to the existing playlists, like `is_local`, then if that is true, this playlist has additional functionality like this but needs to be excluded from the reindex task, if that is false, this playlist is a regular youtube playlist.

With that you can reuse a lot of the existing functionality and don't end up duplicating everything.

I hope that makes sense.

greg321321:
Your approach makes sense, I'll take a look.

bbilly1:
feel free to reopen with the new approach.

greg321321:
ok

PR #620, opened 2023-12-01
Feature 590 custom playlist

greg321321:
tracking issue: https://github.com/tubearchivist/tubearchivist/issues/590

This pull request is ready for review. 

The features are:
1. add/remove custom playlist
2. add/remove video to/from custom playlist
3. move video on custom playlist


tubearchivist/home/src/ta/users.py
```diff
@@ -21,6 +21,7 @@ class UserConfigType(TypedDict, total=False):
     view_style_channel: str
     view_style_downloads: str
     view_style_playlist: str
+    view_style_custom_playlist: str
```
greg321321, line 24: don't need will remove

tubearchivist/home/templates/home/custom_playlist_id.html
```diff
@@ -0,0 +1,102 @@
+{% extends "home/base_playlist.html" %}
```
greg321321, line 1: don't need this file will remove

bbilly1:
At first glance, I think that goes into the right direction. I'll take a closer look later. I know, we should have a better logging setup than just simple print statements, but for now let's keep this PR contained to the custom playlists functionality only. 

You also might need to rebase on testing branch, it looks like you are behind.

greg321321:
lint issues resolved. my version of pyflakes is reporting errors on code that I did not change. Perhaps the pipeline pyflakes will be ok, but i cannot verify.

bbilly1:
This now also does audio channels? How come? Please undo that, so I can review these two PRs independently as that functionality isn't really depended on each other.

greg321321:
> This now also does audio channels? How come? Please undo that, so I can review these two PRs independently as that functionality isn't really depended on each other.

agree, that's weird, you guys were not meant to get that commit. I was doing some integration testing on my own while waiting for this to get into the testing branch. I created a child branch of one of the features in my forked repo, then pulled the changes from the other feature into the child branch. I don't understand why the commits from this child branch got into this PR. Who knows, maybe I setup the branch wrong for github. We use gitlab at work, this type of technique works ok there. I'll update.

greg321321:
reverted to [9c22bfb](https://github.com/tubearchivist/tubearchivist/pull/620/commits/9c22bfb7f887770d37069aee1b52664e784401b8)

should be ok now.

bbilly1:
OK, took a close look at this, run it, created some playlists, I really like it. :-)

Few comments though:
- With our current frontend limitation, please move the functionality to add videos to playlists to the video detail page. This removes a lot of code duplication, will reduce future maintenance burden if we ever want to change something there. This also makes the interface shift less prominent if you have like 20+ custom playlists. You know, we are datahoarders here, you already know someone will create 100 custom playlists and post a screenshot on discord. :-)
- I like the thumbnail implementation, to imitate the behavior from YT is a great choice. But you can avoid the call out to YT to download the thumbnail as that video will already have the thumb cached and you should be able to just copy that to the correct location. Thumb path building is implemented [here](https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/home/src/download/thumbnails.py#L141), you should be able to reuse that, this will also remove the need to use a celery task, further reducing complexity. 
- When creating local playlists, create them as `{"playlist_active": false}` so we can avoid the unneeded try to fetch that from YT with the reindex task.
- For the UUID building, let's prefix that with `TA_` so we know it's a locally created item, so we have a way to separate them again if we need to. Plus we can apply same logic to videos and channels too later.

greg321321:
will update. I do like the idea of having this '...' menu wherever I see a video in the UI. Reason is, I like to 'grab and go' the content. The use case is: I see an interesting video on the main page or search results or channel listing, but don't want to watch it now. So I quickly add it to a 'favorites' playlist. I use the playlist to watch my bookmarked videos. Having the menu item in the details page requires an extra click and an extra page load.

Second point is if the only place we allow the 'add video to playlist' function is in the video details page, that page (at least on my browser) auto starts the video content. This video ends up on the main page 'Continue Watching' list. This section of the UI will get cluttered with videos when I add them to a playlist. I was thinking of a new feature where if the user toggles the watched checkbox to false, it would clear out watch history from the video, and video can disappear from my already bussy 'Continue Watching' section.

If you are concerned about the UI sliding down, I can propose a popup modal with a scroll pane. I can also consolidate the code so it is more efficient to maintain.

bbilly1:
I really want to reduce complexity and code duplication whenever possible. We'll revisit the dots menu thing again, once we have a more sustainable frontend solution in place than what we have now. So please stick with having the playlist menu on the video detail page so we don't add to the current problem.

And yes, you are right with the autoplay thing, I thought I had dealt with that before, but I just didn't realize that this is still set, my Firefox automatically disables that. Also with the continue watching, I've been meaning to skip sending progress before the 10 seconds mark, or something. This should deal with both of these things: https://github.com/tubearchivist/tubearchivist/commit/8778546577d8e87df7f1317436d9e75202a69fd0

greg321321:
Hi bbilly1 can you provide me write access to testing branch? I tried to rebase my branch and "undo/redo" my changes on reindex.py. Neither approach avoided the merge conflict.

bbilly1:
Rebase us usually not sufficient when you have a merge conflict that needs manual interaction to resolve. Usual approach is to merge the upstream branch into your local branch, then resolve the conflicts before committing, then it should be OK to merge your feature back into upstream after that.

So more or less: pull `testing` branch locally to make sure you got the newest changes, then checkout your feature branch, then merge testing branch into your feature branch, resolve all conflicts, and push the merge commit. 

I gave it a shot, please double check to make sure I did it correctly: https://github.com/tubearchivist/tubearchivist/pull/620/commits/8dadaa4979d08d8f21e5eae1984d2382b7e916fd.

greg321321:
thanks ok, I will know for next time. The commit looks ok.

tubearchivist/home/views.py
```diff
@@ -759,13 +760,16 @@ def _update_view_data(self, playlist_id, playlist_info):
             + "{return params.scores[doc['youtube_id'].value];} "
             + "return 100000;"
         )
+        if playlist_info["playlist_type"] == "custom":
```
bbilly1, line 764: I'm not following here: Why is that addition needed here? The original query was matching against the playlist id, custom playlists should be able to do that aswell. Or am I missing something?

bbilly1 (commented):
OK, took a detailed look at it, I pushed a few additional changes, validating things in the API, and a few other mostly a matter of preference changes.

Then I've added some questions where I couldn't follow the reasoning, maybe you can clarify?

Then we are ready to merge. :-)

tubearchivist/home/src/index/channel.py
```diff
@@ -213,7 +213,13 @@ def delete_playlists(self):
         all_playlists = self.get_indexed_playlists()
         for playlist in all_playlists:
             playlist_id = playlist["playlist_id"]
-            YoutubePlaylist(playlist_id).delete_metadata()
+            playlist = YoutubePlaylist(playlist_id)
```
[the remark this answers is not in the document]
greg321321: I think you are correct here and the code can be removed. I don't think a custom playlist would ever show up in a channel list of playlists. I searched the codebase for playlist_channel_id. I see it is set to "None" for custom playlists, so this code path would never get executed.

tubearchivist/home/views.py
greg321321: you may be correct. When we change the position of a video in a custom playlist, we always re-set the idx attribute of all videos in the playlist. So the original logic in this function should be ok. I think I had a previous implementation where I was relying on the order of the videos in the stored array. Using idx to identify the order of videos is more consistent to what is already coded. So perhaps this is a leftover of the previous thinking.

tubearchivist/home/src/index/channel.py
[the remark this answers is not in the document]
bbilly1: very well, I undid the changes.

tubearchivist/home/views.py
bbilly1: sounds good, undid the changes.

bbilly1:
OK, then let's move forward with that. We can always keep iterating from here.

Issue #610, opened 2023-11-21
[Feature Request]: Add Copy button for API Key


stevenhorner:
### I've read the documentation

- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [feature request](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#feature-request) section.

## Is your feature request related to a problem? Please describe.
Not a bug or a problem, but it can be hard to select the API key with a laptop trackpad or on a phone and the Revoke button is quite close and easy to press accidently. Presumably the button asks to confirm anyway, didn't test this due to not wanting to revoke my current keys.

## Describe the solution you'd like
Add a simple Copy button that allows copying the API key to the clipboard and/or put the key in a text box that makes it slightly easier to select.

### Your help is needed!

- [ ] Yes I will work on this in the next few days or weeks.

bbilly1:
Some more explanations in the linked PR, but as it looks like, this can't be reliably implemented without SSL, I'm closing this as won't fix.

PR #633, opened 2023-12-15
 [Feature Request]: Add Copy button for API Key #610

smitchell6879:
This adds the copy button. Works for me on local pc using firefox and brave browser on android. Should work in the future with https also if the documentation I read is correct. The was it is now it will fail the https and fall back to the http version.

### DIFF
--- a/tubearchivist/home/templates/home/settings_application.html
+++ b/tubearchivist/home/templates/home/settings_application.html
@@ -155,7 +155,8 @@
         <div class="settings-item">
             <p>API token: <button type="button" onclick="textReveal(this)" id="text-reveal-button">Show</button></p>
             <div id="text-reveal" class="description-text">
-                <p>{{ api_token }}</p>
+                <p id="text-token">{{ api_token }}</p>
+                <button type="button" onclick="textCopy()" id="text-copy-button">Copy</button>
                 <button class="danger-button" type="button" onclick="resetToken()">Revoke</button>
             </div>
         </div>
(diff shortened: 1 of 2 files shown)


bbilly1:
Super unfortunate, I wasn't aware that `navigator.clipboard` is only available on https. Looks like `execCommand()` has been depreciated for 3y+ now. If I'd have been aware of that, I'd have closed that issue as *won't fix*.

Honestly I don't think we should add functionality to this project, that can't be implemented without introducing depreciated code.

I did quite some research, but as far as I can see, these two options are the only way to do this, or am I missing something?

smitchell6879:
Nope that is what I had found also. I haven't looked at what server your using to host the webpage yet but I know I had used uvicorn on one of my old projects and you had to choose on start to run either http or https but if you did use https you could create your own certificates. Of course doing this would make it so the first time the user went to the login page they would have to accept the risk to continue as the certificates where not from an authority since I had created them.  But in theory at that point you could use this copy function.

bbilly1:
Thank you for looking at this, but it looks like that currently can't be implemented easily. 

In my opinion, SSL should be handled at reverse proxy level, and not application level, so we can't make assumptions about that. This currently uses uwsgi, but hopefully, some time down the road, we'll switch to an asgi server, to be able to use those sweet sweet websockets. 

But for the moment, as that is just a convenience feature, I'd rather not introduce something that uses depreciated methods.

Issue #500, opened 2023-07-22
[Bug]: Subscription Rescan Memory Useage


PhuriousGeorge:
### I've read the documentation

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

Unraid

## Describe the bug
I honestly can't understand why TA gobbles RAM so much.  I've 'only' got 47GB available to the server and recently TA has been overloading & locking up my server if I have any other container(s) cumulatively using more than 5GB RAM, requiring a manual reboot.  I've discussed this a few times and received various answers before I ultimately decided to limit the container.  I've limited TA to 8GB using `--memory=8G`. This is not related to ES at all, TA is the only container I've had to limit.  Now with this limit implemented, my subscriptions completely fail to scan every time.  I receive a "lost worker" error in the log, which I can only guess is the rescan task, as the rescan has not completed in the last 4 days.

I do understand I'm an "edge case", but the resource usage is a bit edgier ;)

## Steps To Reproduce
Have 1944 subscriptions
Limit TA container to 8GB RAM
Initiate 'Rescan Subscriptions' task
Receive "Worker Lost" message

### Relevant log output

```shell
[2023-07-21 20:23:13,077: ERROR/MainProcess] Process 'ForkPoolWorker-32' pid:115 exited with 'signal 9 (SIGKILL)'
[2023-07-21 20:23:13,093: ERROR/MainProcess] Task handler raised error: WorkerLostError('Worker exited prematurely: signal 9 (SIGKILL) Job: 0.')
Traceback (most recent call last):
  File "/root/.local/lib/python3.10/site-packages/billiard/pool.py", line 1265, in mark_as_worker_lost
    raise WorkerLostError(
billiard.exceptions.WorkerLostError: Worker exited prematurely: signal 9 (SIGKILL) Job: 0.
```

bbilly1:
[...]

PR #639, opened 2023-12-29
Limit worker lifespan - RAM useage mitigation

PhuriousGeorge:
Limit worker lifespan to save our precious RAM as discussed on [Discord](https://discord.com/channels/920056098122248193/1179480913701241002/1180026088802496512)

Mitigates #500 though RAM usage can still ramp rather high before worker is culled

### DIFF
--- a/docker_assets/run.sh
+++ b/docker_assets/run.sh
@@ -18,7 +18,7 @@ python manage.py ta_migpath
 
 # start all tasks
 nginx &
-celery -A home.tasks worker --loglevel=INFO &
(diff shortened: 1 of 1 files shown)


lamusmaser:
First note, let's marge this against tubearchivist:test, instead of tubearchivist:master. Should be able to change that on your side @PhuriousGeorge. Otherwise, this looks good to me and includes only what is requested.

PhuriousGeorge:
I have no idea how to change branches with a pull request once summitted so I'll close this and created #644

PR #654, opened 2024-01-21
Update comments.py

danieljue:
Cleaned up logic, based on what I interpreted from the code: 

Only try to grab new comments from youtube if ES was empty.

Fixed an unexecuted branch compare.
The return inside "if not self.comments_format and es_comments["comment_comments"]:" could never be reached because of the comparison above it.

Fixed a potential "TypeError: 'bool' object is not subscriptable" due to indexing on es_comments when it is False.

By reducing the number of wasteful calls we can further escape throttling.   (Although if the intention is to force replacement of old comments with any new ones, then why even query ES?)

### DIFF
--- a/tubearchivist/home/src/index/comments.py
+++ b/tubearchivist/home/src/index/comments.py
@@ -164,18 +164,16 @@ class Comments:
         self.check_config()
         if not self.is_activated:
             return
-
-        self.build_json()
-        if not self.json_data:
-            return
-
+        # check for existing comments first (could return False).
         es_comments = self.get_es_comments()
-
-        if not self.comments_format:
+        if es_comments and es_comments["comment_comments"]:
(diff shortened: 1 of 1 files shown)


bbilly1:
Thinking here was, that same functionality can be used for indexing new videos *and* reindexing existing videos and refresh the comments.

The call to ES is to catch cases where for example the channel owner removes comments after a certain time for whatever reason but you have it indexed locally, so in that case we don't want to delete the already indexed comments.

danieljue:
Oh ok, I'll have to think about that use case, I didn't even consider the channel owner removing comments!   I can defintely understand merging of functionality into one method.  I still don't see how line 177's second comparison can be reached, since 174 would cause an early return, and 177 starts with the same comparison as 174.

PR #660, opened 2024-01-29
ES index mapping: numbers should be numbers

peterfromthehill:
Hi,

the number of replicas should be specified as numbers and not as strings.
it is more compatible with other solutions like zincsearch.

best regargs

### DIFF
--- a/tubearchivist/home/src/es/index_mapping.json
+++ b/tubearchivist/home/src/es/index_mapping.json
@@ -8,7 +8,7 @@
                 }
             },
             "expected_set": {
-                "number_of_replicas": "0"
+                "number_of_replicas": 0
             }
         },
         {
@@ -88,7 +88,7 @@
                         }
                     }
                 },
-                "number_of_replicas": "0"
+                "number_of_replicas": 0
             }
         },
         {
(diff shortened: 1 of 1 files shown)


bbilly1:
Unfortunately Elasticsearch returns that value as a string and not as an int. You can test it easily, create a testing index, e.g.:
```
PUT testing
```
Then get the settings:
```
GET testing/_settings
``` 
and you'll get:
```json
{
  "testing": {
    "settings": {
      "index": {
        "routing": {
          "allocation": {
            "include": {
              "_tier_preference": "data_content"
            }
          }
        },
        "number_of_shards": "1",
        "provided_name": "testing",
        "creation_date": "1707162826292",
        "number_of_replicas": "1",
        "uuid": "LvVv3U6TS6-Nx6eB9vQ40w",
        "version": {
          "created": "8500008"
        }
      }
    }
  }
}
```
where ES returns `number_of_replicas` as a string, even if their docs says you should add it as an int [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html), adding it as a string will still work...

At application startup TA compares the expected mappings and settings with the actual mappings and settings. If you test your changes you suggested in this PR, you'll realize that this will rebuild the index at every startup as actual will never be equal to expected. So that's not something we can do.

PR #669, opened 2024-03-01
Log cookiejar.LoadError from POST /api/cookie/

Gentoli:
Help identify the root cause for https://github.com/tubearchivist/browser-extension/issues/22. 

This would log:
```
cookie file is invalid: invalid Netscape format cookies file <_io.StringIO object at 0x7f5793dd7490>: 'payments.youtube.com\tTRUE\t/\tFALSE\t1711779732\tOTZ\t<some-value>'
```
Note: it could log out the line containing the cookie (secret) that's causing the error.


bbilly1:
looks good, thanks for adding this here.

Issue #634, opened 2023-12-16
[Bug]: Adding a playlist to the download queue doesn't work if the playlist has been subscribed to before


zombachu:
### I've read the documentation

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

Synology

## Describe the bug
I had a playlist I was subscribed to but was behind in archiving by a couple months, so I wanted to sync it all at once by adding it to the download queue. However, none of the missing videos were added and the playlist unsubscribed itself and marked itself as deactivated too. Logs show TubeArchivist was able to get the playlist from the url I provided but it seemed to only check the videos it already had. Resubscribing to the playlist and doing rescan subscriptions didn't pick up the unarchived videos either. I'm able to add the videos individually manually so I don't think it's an access issue but I really really do not want to need to do this.

## Steps To Reproduce
1. Have a subscribed playlist in TubeArchivist that is not fully archived
2. Add the playlist to the download queue
3. sadness

## Expected behavior
The unarchived videos to be added to the download queue

### Relevant log output

```shell
[pid: 3297|app: 0|req: 5593/5593] 172.17.0.1 () {52 vars in 1275 bytes} [Sat Dec 16 01:17:20 2023] POST /api/download/ => generated 104 bytes in 34 msecs (HTTP/1.1 400) 8 headers in 279 bytes (1 switches on core 0)
[2023-12-16 01:18:05,372: INFO/MainProcess] Task extract_download[f8510c88-5047-4262-ad0b-fe3c9ea0c60c] received
[1;33m[2023-12-16 01:18:05,645: WARNING/ForkPoolWorker-2] extract_download create callback[0m
[1;33m[2023-12-16 01:18:19,162: WARNING/ForkPoolWorker-2] PLiasgr1hOV7crPuf1h0AzZOFUQqtOcI5H: get metadata from es[0m
[1;33m[2023-12-16 01:18:19,434: WARNING/ForkPoolWorker-2] K7_XGfcytPQ: skipped adding already indexed video to download.[0m
[...]

PR #670, opened 2024-03-03
fix: playlist not getting updated after first index

svuong-dev:
Addresses the bug #634 
inside playlist.py
```
    def build_json(self, scrape=False):
...
        if scrape or not self.json_data:
```
self.json_data will only be empty if the playlist has never been indexed. Since `scrape` defaults to `False`, both conditions mean subsequent playlist downloads will never do any scraping. Adding `scrape=True` to the parent call fixes this scenario (tested on my own setup). 

Note that playlists don't seem to update via subscriptions scans still. I suspect we need a similar fix there, but I don't have the bandwidth to dig into that atm.

### DIFF
--- a/tubearchivist/home/src/download/queue.py
+++ b/tubearchivist/home/src/download/queue.py
@@ -227,7 +227,7 @@ class PendingList(PendingIndex):
     def _parse_playlist(self, url):
         """add all videos of playlist to list"""
         playlist = YoutubePlaylist(url)
-        playlist.build_json()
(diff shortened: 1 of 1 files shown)


bbilly1:
This needs more to look into. That flag is there to avoid repeated unneeded refreshes out to YT, as that method might get called multiple times when adding multiple videos to the same Playlist.

The whole Playlist integration will deserve some refactoring. That didn't see much improvements since first implementation.

Issue #286, opened 2022-07-27
Deleted videos should be added to the ignore list


samdoshi:
It would be nice if videos (from subscribed channels) that have been deleted don't show up again in the download queue. Perhaps they could be added to the ignore list automatically when they are deleted?

bbilly1:
Yeah, we had a short conversation before about that. I was a bit hesitant, if that happens automatically, you won't be able to download it again, until you go to the ignored list and delete it from there. So that could be confusing, as any ignored videos will skip without any warning even if you add it explicitly. 

But maybe there is a way how this could be done, so that it's obvious? Maybe as part of the delete confirmation? When you click on *Delete Video* you get *Delete*, *Delete and ignore* and *cancel*?

samdoshi:
I think you've got four options:

 1. Always add it to the ignore list
 1. An option in the settings to "add deleted videos to the ignore list"
 1. Your suggestion "Delete, Delete and ignore and cancel"
 1. Similarly to your suggestion, when you click delete you get the options "Delete", "Cancel", but also a checkbox "Also add to ignore list". The state of the checkbox can be remembered between uses.

I think I would like the 2nd or 4th options.

I think with options 1 and 2, you just need to make sure that there is a message in the confirmation saying that the deleted video will be added to the ignore list as well, so that users know where to find it again.

Joshfindit:
I would like to suggest adding a flag to the video metadata so that that flag can be used to ignore/hide in the UI.

This is related, but I actually want to be able to view all videos I have that have been removed from YouTube (and check if there are any freshly removed videos)

Something like:`removed = user_deleted|uploader_made_private|uploader_deleted`

cmuc24:
oh nice, looks like similar to my suggested blacklist feature [last year ](https://github.com/tubearchivist/tubearchivist/discussions/110) 
thumbs up for that 👍

SmartPhoneLover:
> I would like to suggest adding a flag to the video metadata so that that flag can be used to ignore/hide in the UI.
> 
> This is related, but I actually want to be able to view all videos I have that have been removed from YouTube (and check if there are any freshly removed videos)
> 
> Something like:`removed = user_deleted|uploader_made_private|uploader_deleted`

Maybe add both options, I mean:
- Delete / Delete & Ignore / Cancel (will free up space on disk)
- Hide / Cancel (will keep the video on disk, but it will be hidden)

[...]

PR #671, opened 2024-03-04
Add option to ignore videos when deleting

jonvmey:
Adds a "Delete and Ignore" option when deleting a video which automatically marks it as ignored in the download queue.

Happy to modify the changes if there's a better way to handle this.

Changes related to #286

### DIFF
--- a/tubearchivist/static/script.js
+++ b/tubearchivist/static/script.js
@@ -381,6 +381,19 @@ function deleteVideo(button) {
   }, 1000);
 }
 
+function deleteAndIgnoreVideo(button) {
+  let to_delete = button.getAttribute('data-id');
+  let to_redirect = button.getAttribute('data-redirect');
+  let endpointDelete = '/api/video/' + to_delete + '/';
+  apiRequest(endpointDelete, 'DELETE');
+  let endpointDownload = '/api/download/';
+  apiRequest(endpointDownload, 'POST', { data: [{ youtube_id: to_delete, status: 'ignore' }] });
+  setTimeout(function () {
+    let redirect = '/channel/' + to_redirect;
+    window.location.replace(redirect);
+  }, 1000);
+}
+
 function deleteChannel(button) {
   let to_delete = button.getAttribute('data-id');
   let apiEndpoint = '/api/channel/' + to_delete + '/';
--- a/tubearchivist/api/views.py
+++ b/tubearchivist/api/views.py
@@ -10,7 +10,7 @@ from api.src.aggs import (
     WatchProgress,
 )
 from api.src.search_processor import SearchProcess
-from home.src.download.queue import PendingInteract
+from home.src.download.queue import PendingInteract, PendingList
 from home.src.download.subscriptions import (
     ChannelSubscription,
     PlaylistSubscription,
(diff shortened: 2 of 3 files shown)


tubearchivist/static/script.js
```diff
@@ -381,6 +381,19 @@ function deleteVideo(button) {
   }, 1000);
 }
 
+function deleteAndIgnoreVideo(button) {
+  let to_delete = button.getAttribute('data-id');
+  let to_redirect = button.getAttribute('data-redirect');
+  let endpointDelete = '/api/video/' + to_delete + '/';
+  apiRequest(endpointDelete, 'DELETE');
+  let endpoingDownload = '/api/download/';
```
barrelltitor, line 389: Here and on the line below shouldn't it be endpoin**t**Download instead of endpoing?

jonvmey: It definitely should.

bbilly1:
Thanks for taking a stab at this. You can even further reuse existing functionality by slightly tweaking the `extrac_dl` task to also take a `status` kwarg, like so:

```python
@shared_task(name="extract_download", bind=True, base=BaseTask)
def extrac_dl(self, youtube_ids, auto_start=False, status="pending"):
    """parse list passed and add to pending"""
    TaskManager().init(self)
    if isinstance(youtube_ids, str):
        to_add = Parser(youtube_ids).parse()
    else:
        to_add = youtube_ids

    pending_handler = PendingList(youtube_ids=to_add, task=self)
    pending_handler.parse_url_list()
    pending_handler.add_to_pending(status=status, auto_start=auto_start)

    if auto_start:
        download_pending.delay(auto_only=True)
```
Then `status` is already implemented in `add_to_pending`. Additionally by handing that off to the task, that can be done async, as that will take a few seconds because it needs to go out to YT to get the data first.

Then from the API side, this would make more sense to implement through the Download Item view `/api/download/<video_id>/` and not the list view, as you are interacting with a single video item, and not with a list. I would suggest to implement this through another `valid_status`, maybe `ignore-force`, to force ignore it even if it doesn't exist? So you can easily separate it with the regular ignore functionality that doesn't need to be called through a task because it's already in the queue.

So partial view something like this:
```python
class DownloadApiView(ApiBaseView):

    valid_status = ["pending", "ignore", "ignore-force", "priority"]

    def post(self, request, video_id):
        """post to video to change status"""
        item_status = request.data.get("status")
        if item_status not in self.valid_status:
            message = f"{video_id}: invalid status {item_status}"
            print(message)
            return Response({"message": message}, status=400)

        if item_status == "ignore-force":
            extrac_dl.delay(video_id, status="ignore")
            message = f"{video_id}: set status to ignore"
            return Response(request.data)

        # rest of the view...

``` 
Then on button click, this would mean two API calls, first to DELETE the video, then POST to `ignore-force` it on the queue, something like this:

```js
function deleteVideo(button) {
  let to_delete = button.getAttribute('data-id');
  let to_ignore = button.getAttribute('data-ignore');
  let to_redirect = button.getAttribute('data-redirect');
  let apiDeleteEndpoint = '/api/video/' + to_delete + '/';
  apiRequest(apiDeleteEndpoint, 'DELETE');

  if (to_ignore !== null) {
    let apiIgnoreEndpoint = '/api/download/' + to_delete + '/';
    apiRequest(apiIgnoreEndpoint, 'POST', { status: 'ignore-force' });
  }

  setTimeout(function () {
    let redirect = '/channel/' + to_redirect;
    window.location.replace(redirect);
  }, 1000);
}
```
While the button would look like this:
```html
<button class="danger-button" onclick="deleteVideo(this)" data-id="{{ video.youtube_id }}" data-ignore data-redirect = "{{ video.channel.channel_id }}">Delete and ignore</button>
```

Does that make sense? Or did I miss something?

Are you up to make these modifications? Otherwise no problem, let me know.

jonvmey:
Thanks for the feedback. I'll look over your suggestions and work on implementing them.

bbilly1:
are you still looking into this? Otherwise let me know.

bbilly1:
closing this in favor of: https://github.com/tubearchivist/tubearchivist/commit/852abf254d0dac5c5b4265df3285218559a943ba

PR #674, opened 2024-03-08
Automatic reloading of uwsgi in docker developer environment

dot-mike:
This makes uwsgi reload when changes are detected for python or static files in Docker environment for developers.
The user only has to refresh the application in the web-browser to see the changes.

I've updated `contributing.md` and added a Makefile to make life easier and I hope you accept it.

Run this command from the terminal to start the improved developer stack or use `make dev` :)
```bash
docker compose -f docker-compose.dev.yml up --build
```


bbilly1:
Honestly this is too much complexity. The more we differ the dev setup to what users are running, the more difficult it is to test things, especially until we have that automated.

Also I want to avoid duplicating the compose file and the uwsgi file.

Rebuilding on change hasn't been too much of an issue for me, it just takes a few seconds, as all layers except for the application can be cached.

Djangos builtin server has autoreload implemented, I've recently started using that. 

More or less in run.sh:
```bash
if [[ -z "$TESTING_ENV" ]]; then
    uwsgi --ini uwsgi.ini
else
    python manage.py runserver 0.0.0.0:8001
fi
```
So when you add `TESTING_ENV` to the compose file this will start the django server instead on port 8001. then you can mount the `/app` folder as a volume and you have autoreload implemented. 

That would be easy enough to explain in CONTRIBUTING.md without needing to duplicate everything and adding overhead.

bbilly1:
we will revisit a better dev setup with hopefully hot reloading when we make the split with the frontend.

PR #676, opened 2024-03-09
Fix comment author is None during comment download

dot-mike:
This fixes a rare edge-case that may happen during parsing of comments generated by Yt-Dlp where author is None. So we default to author_id. This can happen mostly for older YT profiles that has not set-up a @-handle.


bbilly1:
Do you have a link where this occurred? I was under the assumption that YT gave everybody a @ handler, with some random number after the channel name.

bbilly1:
OK, found it, strange, but easy enough work around. Just remove the whitespace for the linter there on 117 and we should be good.

PR #682, opened 2024-03-14
Add additional user scripts.

lamusmaser:
Adding additional user scripts; two of my repos and one of another, highly referenced repo.


bbilly1:
Hmm, unfortunate that this symlink repo doesn't have a license, making it not open source. How should we handle this?

Maybe a more clearer indication that "this is just a collection of links" and "see license info in the respective repo"?

bbilly1:
I added some clarifications...

PR #690, opened 2024-03-26
Add support for connecting to Redis via unix socket

christoph-heiss:
Hi!

Pretty much as it says on the tin.
This adds support for a `REDIS_SOCKET` environment variable, which - if set - is used to connect to Redis instead of host + port.

The "priorization" is done by the `Redis` constructor itself, which prefers `unix_socket_path` over `host` and `port`.

A second commit (438d11b58e099f165408f8dabeb462c7d2a66d34) is included, which is a small addition to print the exception message when the connect to redis fails. Helpful to troubleshoot what is going on.

- [x] branch is based on `testing`
- [x] ran `./deploy.sh validate` on each modified file

Thanks for considering!

### DIFF
--- a/tubearchivist/home/src/ta/config.py
+++ b/tubearchivist/home/src/ta/config.py
@@ -47,8 +47,8 @@ class AppConfig:
 
                 return config
 
-            except Exception:  # pylint: disable=broad-except
-                print(f"... Redis connection failed, retry [{i}/10]")
+            except Exception as e:  # pylint: disable=broad-except
+                print(f"... Redis connection failed, retry [{i}/10]: {e}")
                 sleep(3)
 
         raise ConnectionError("failed to connect to redis")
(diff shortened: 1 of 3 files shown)


Iliannnn:
Very nice addition. Thank you!

bbilly1:
Hmm, not sure what the benefit is to use sockets over regular TCP. Why would you want that additional complexity? Host and Port is much simpler and I think much more common in a selfhosted docker compose environment as you naturally need network communication between containers.

I generally try to avoid adding unnecessary complexity, especially if the current implementation is working fine.

bbilly1:
This is a docker application, docker is the only supported installation method, but obviously if you know a little bit how things work, it's not too complicated to install this on bare metal.

Nonetheless, first concern with adding new features is to reduce complexity, or at least not add to it unnecessarily, even more so, you can easily use redis over TCP and sockets is not a requirement, even for an unsupported installation setup like yours.

Maintainability is first priority, and I don't want to introduce environment variables only used by possible one person. Sorry, I hope you can understand.

PR #628, opened 2023-12-10
DELETE ME - Feature #623 channel audio mode

greg321321:
Issue #623 

Works ok with AntennaPod and PodcastAddict Android apps. 

Ready for review.

PR #702, opened 2024-04-27
Add Page Size Overrides per Channel

Boo1098:
This is a minimum viable product for adding per channel overwrites on the number of videos, shorts, and livestreams to query when refreshing subscriptions. Tested all 3 overrides and they worked. The current method of resetting the override is clunky (setting to negative number). Open to ideas on how to better implement that. I've also upended some of the build query in subscriptions and haven't fully tested if that messes with things, but have been running this on my install the past few days with no ill effects that I've found.

New channel -> about overwrites page:
image

Sorry I can't contribute to the burn down of maintainability items, I'm not too experienced and don't know where to start for any of the remaining open items.

### DIFF
--- a/tubearchivist/home/src/download/subscriptions.py
+++ b/tubearchivist/home/src/download/subscriptions.py
@@ -38,11 +38,15 @@ class ChannelSubscription:
         return all_channels
 
     def get_last_youtube_videos(
-        self, channel_id, limit=True, query_filter=VideoTypeEnum.UNKNOWN
+        self,
+        channel_id,
+        limit=True,
+        query_filter=None,
+        channel_overwrites=None,
     ):
         """get a list of last videos from channel"""
-        queries = self._build_queries(query_filter, limit)
-
+        query_handler = VideoQueryBuilder(self.config, channel_overwrites)
+        queries = query_handler.build_queries(query_filter)
         last_videos = []
 
         for vid_type_enum, limit_amount in queries:
@@ -50,54 +54,24 @@ class ChannelSubscription:
                 "skip_download": True,
                 "extract_flat": True,
             }
+            vid_type = vid_type_enum.value
+
             if limit:
                 obs["playlistend"] = limit_amount
 
-            vid_type = vid_type_enum.value
-            channel = YtWrap(obs, self.config).extract(
-                f"https://www.youtube.com/channel/{channel_id}/{vid_type}"
-            )
-            if not channel:
+            url = f"https://www.youtube.com/channel/{channel_id}/{vid_type}"
+            channel_query = YtWrap(obs, self.config).extract(url)
+            if not channel_query:
                 continue
+
             last_videos.extend(
-                [(i["id"], i["title"], vid_type) for i in channel["entries"]]
+                [
+                    (i["id"], i["title"], vid_type)
+                    for i in channel_query["entries"]
+                ]
             )
 
         return last_videos
-
-    def _build_queries(self, query_filter, limit):
-        """build query list for vid_type"""
-        limit_map = {
-            "videos": self.config["subscriptions"]["channel_size"],
-            "streams": self.config["subscriptions"]["live_channel_size"],
-            "shorts": self.config["subscriptions"]["shorts_channel_size"],
-        }
-
-        queries = []
-
-        if query_filter and query_filter.value != "unknown":
-            if limit:
-                query_limit = limit_map.get(query_filter.value)
-            else:
-                query_limit = False
-
-            queries.append((query_filter, query_limit))
-
-            return queries
-
-        for query_item, default_limit in limit_map.items():
-            if not default_limit:
-                # is deactivated in config
-                continue
-
-            if limit:
-                query_limit = default_limit
-            else:
-                query_limit = False
-
-            queries.append((VideoTypeEnum(query_item), query_limit))
-
-        return queries
 
     def find_missing(self):
         """add missing videos from subscribed channels to pending"""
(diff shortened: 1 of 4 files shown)


bbilly1:
Thanks for taking a stab at that. 

Current backlog:
- Schedule Migration: https://github.com/tubearchivist/tubearchivist/tree/feature-beat-model, almost ready to merge
- Channel audio mode: https://github.com/tubearchivist/tubearchivist/pull/628, not yet looked at
- Then this.

Please be patient.

bbilly1:
OK, I took some time to look into this. That is some old code, I don't think I ever touched it since project beginning. :-)

I think it's time to break out the query building into a separate class there. That is more or less self contained. Then this also prepares us for future expansion if we ever need to add additional types. 

This is what I came up with:

```python
class VideoQueryBuilder:
    """Build queries for yt-dlp."""

    def __init__(self, config: dict, channel_overwrites: dict | None = None):
        self.config = config
        self.channel_overwrites = channel_overwrites or {}

    def build_queries(
        self, video_type: VideoTypeEnum | None, limit: bool = True
    ) -> list[tuple[VideoTypeEnum, int | None]]:
        """Build queries for all or specific video type."""
        query_methods = {
            VideoTypeEnum.VIDEOS: self.videos_query,
            VideoTypeEnum.STREAMS: self.streams_query,
            VideoTypeEnum.SHORTS: self.shorts_query,
        }

        if video_type:
            # build query for specific type
            query_method = query_methods.get(video_type)
            if query_method:
                query = query_method(limit)
                if query[1] != 0:
                    return [query]
                return []

        # Build and return queries for all video types
        queries = []
        for build_query in query_methods.values():
            query = build_query(limit)
            if query[1] != 0:
                queries.append(query)

        return queries

    def videos_query(self, limit: bool) -> tuple[VideoTypeEnum, int | None]:
        """Build query for videos."""
        return self._build_generic_query(
            video_type=VideoTypeEnum.VIDEOS,
            overwrite_key="subscriptions_channel_size",
            config_key="channel_size",
            limit=limit,
        )

    def streams_query(self, limit: bool) -> tuple[VideoTypeEnum, int | None]:
        """Build query for streams."""
        return self._build_generic_query(
            video_type=VideoTypeEnum.STREAMS,
            overwrite_key="subscriptions_live_channel_size",
            config_key="live_channel_size",
            limit=limit,
        )

    def shorts_query(self, limit: bool) -> tuple[VideoTypeEnum, int | None]:
        """Build query for shorts."""
        return self._build_generic_query(
            video_type=VideoTypeEnum.SHORTS,
            overwrite_key="subscriptions_shorts_channel_size",
            config_key="shorts_channel_size",
            limit=limit,
        )

    def _build_generic_query(
        self,
        video_type: VideoTypeEnum,
        overwrite_key: str,
        config_key: str,
        limit: bool,
    ) -> tuple[VideoTypeEnum, int | None]:
        """Generic query for video page scraping."""
        if not limit:
            return (video_type, None)

        if overwrite_key in self.channel_overwrites:
            overwrite = self.channel_overwrites[overwrite_key]
            return (video_type, overwrite)

        if overwrite := self.config["subscriptions"].get(config_key):
            return (video_type, overwrite)

        return (video_type, None)
```

That then should simplify `get_last_youtube_videos` to something like this:
```python
def get_last_youtube_videos(
    self,
    channel_id,
    limit=True,
    query_filter=None,
    channel_overwrites=None,
):
    """get a list of last videos from channel"""
    query_handler = VideoQueryBuilder(self.config, channel_overwrites)
    queries = query_handler.build_queries(query_filter)
    last_videos = []

    for vid_type_enum, limit_amount in queries:
        obs = {
            "skip_download": True,
            "extract_flat": True,
        }
        vid_type = vid_type_enum.value

        if limit:
            obs["playlistend"] = limit_amount

        url = f"https://www.youtube.com/channel/{channel_id}/{vid_type}"
        channel_query = YtWrap(obs, self.config).extract(url)
        if not channel_query:
            continue

        last_videos.extend(
            [
                (i["id"], i["title"], vid_type)
                for i in channel_query["entries"]
            ]
        )

    return last_videos
```

This is minimally tested... Not sure if that covers all cases... But is much more explicit and less confusing to what I had before there and to what you had to work with. :-)

I'm also thinking that we might want to have a more sophisticated download for at some point, e.g. when you add a channel to the form, to download x amount of videos, or something like that...

Does that make sense?

Also you might want to rebase on testing branch, I pushed quite a few things since you branched here.

Boo1098:
Yeah this looks much simpler. I'll put it together into a commit


> I'm also thinking that we might want to have a more sophisticated download for at some point, e.g. when you add a channel to the form, to download x amount of videos, or something like that...
> 
> Does that make sense?

This sounds great. It would fix the problem I had when setting up and importing all my youtube subscriptions where I really only wanted new videos so I had to manually ignore all the added videos. I think that's outside the scope of this PR but I'll keep it in mind.

Boo1098:
Added your code, works great. Only had to make one change [here](https://github.com/tubearchivist/tubearchivist/pull/702/commits/e4c2ac049030bdf682b4c5e2d3fae2c72938140e#diff-640e302a532376d6adf23099169898c213d363f54e06b0affb1e9b32dc360310R206) since in the case the main config entry is False it means no videos of that type should be queried.

Boo1098:
> I also [added a check](https://github.com/Boo1098/tubearchivist/blob/per_channel_shorts/tubearchivist/home/src/download/subscriptions.py#L202) that the channel override is not None as that is what it was set to when it was removed.

Looking into this more, I think this was from me not properly deleting entries from the channel overrides config when I was testing earlier. Fresh install does not need this check as the keys are actually deleted when the override is unset.

bbilly1:
OK, I did some testing:

1. Disable on application level, enable in channel: E.g. in settings videos: 5, streams 0, shorts 0, then enable in channel, e.g. overwrite shorts 5. That works as expected.
2. Enable on application level, disable on channel: E.g. in settings set all to 5, then in channel set shorts to 0 to disable. That works as expected.
3. Enable on application, overwrite on channel: e.g. in settings set all to 5, on channel set all to 10. That works as expected.

Also removing overwrites is working fine too.

All in all, I think this is very nice. :-) Thanks for putting in the effort.

Then UI/UX wise, that will get an overhaul anyways soon, so we can avoid the negative number or zero thing. But with the explanation, I thin that's fine for the moment. 

Am I missing something?

Boo1098:
Sounds good to me! I'm good to merge when you are.

PR #735, opened 2024-05-25
Fix run script respect system shutdown requests

mastoca:
if you're running in Docker and the host needs to restart, the run script & celery workers can slow/hinder reboots & shutdowns.

This change fixes it so system calls are respected inside the shell script.

### DIFF
--- a/docker_assets/run.sh
+++ b/docker_assets/run.sh
@@ -16,8 +16,8 @@ python manage.py ta_connection
 python manage.py ta_startup
 
 # start all tasks
-nginx &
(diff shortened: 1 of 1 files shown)


mastoca:
When you shutdown/reboot the host, (without docker compose down) the uwsgi instances sometimes seem to not get the signal to terminate from the bash script. This will cause a ~60 second (depending on your OS timeouts) delay in the shutdown/restart. These changes fixed the issue as the uwsgi threads get the SIG and power down.

PR #740, opened 2024-06-03
Allow running the container as non-root

farcaller:
This uses a venv at `/applib` instead of the default /root/.local store for the PYTHONPATH so that it's easier to run the container as a non-root user.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential gcc libldap2-dev libsasl2-dev libssl-dev git
 
 # install requirements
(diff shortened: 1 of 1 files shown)


bbilly1:
This is all going to change soon anyways with the refactoring and fronted change. This is not a good time to make changes like that. Running TA as a different user in the container will need more changes too than just where the python libraries go. And will bring a whole set of permission problems that needs solving...

farcaller:
It's your project, so you decide, but just to confirm: this is the only change I had to do to run TA as an unprivileged user. It also significantly reduced the warning noise from uwsgi that seemed really unhappy about running as root.

Issue #623, opened 2023-12-03
[Feature Request]: channel audio mode


greg321321:
### Your Feature Request

Provide an audio mode to channels.

User story #1:
1. As a user, I want to play tube archivist channel content as audio only on my podcast app.

Expected Behavior #1:
1. TA shall provide an "audio/video podcast RSS link" on the TA channel page.
2. TA shall provide an "audio/video podcast RSS link" on the TA playlist page.
3. Upon request of the "audio/video podcast RSS link" data, TA shall provide podcast data in accordance with the [PSP-1: The Podcast RSS Standard](https://github.com/Podcast-Standards-Project/PSP-1-Podcast-RSS-Specification).
4. Upon request for audio media, TA shall provide podcast audio data utilizing ffmpeg on the fly transcoding.
5. TA shall provide podcast audio data in mp3 format.
6. TA podcasts shall be consumable by the following applications: [Antenna Pod](https://www.pcmag.com/picks/the-best-podcast-player-apps).
7. TA shall protect podcast data from unauthorized access.
8. When a podcast app sends more than one simultaneous request for podcast content, TA shall prevent simultaneous requests from negatively impacting the user experience on the main GUI.

Guidance per Simon bbilly:
Step one is to have audio mode, that should be functional in the TA interface, so you can listen to audio only. This would have to be configured per channel, ideally the options would be: video only (currently), audio only, video + audio. This would give all options. Then audio only should download the mp3 directly with yt-dlp, and video + audio could use ffmpeg for post-processing to extract the audio from the video file. 

### Your help is needed!

- [X] Yes I will work on this in the next few days or weeks.

bbilly1:
There are many popular podcasts out there that are pretty much talk shows, they have an interview that is filmed, but they are also a podcast, as in you don't really need the video to follow the conversation, but you might want to. So you might want to get the audio version in your podcast app, but also have the video version to watch the interview. I've also learned that YT has a separate /podcasts section:

Examples:
https://www.youtube.com/@JupiterBroadcasting/podcasts
https://www.youtube.com/@LinusTechTips/podcasts
https://www.youtube.com/@H3Podcast/podcasts
https://www.youtube.com/@ThroughTheWeb/podcasts

But some research also showed, that most podcasts are playlists and not channels, or group of playlists in a channel. So the previous thinking of having audio channels is still valid for music channels, but not really for podcasts and talkshows.

[...]

PR #747, opened 2024-06-13
Feature #623 channel audio mode

greg321321:
Issue #623 .


greg321321:
ready for review

bbilly1:
Sorry, I should have looked closer at this earlier in the process.

Honestly, this is way too complicated. Also many of these things will change with the new frontend.

Sorry this is best to but on hold for now. For the sake of reference, things I'd do differently:

- Auth like that in the url is very insecure, as that token, possibly with read/write access will be in plain text in the podcast app, in the logs, probably more places. We'd be better off to remove auth read only on the media files and the XML file, similar as with the cast integration. Or look into other ways how to lock down access...
- The caching approach and transcoding the media files on request will likely not be feasible on lower end hardware, as you'll reach timeouts. So mp3 files should just go next to the mp4 files. Then you also don't need a separate mp3 route for that, as it's just part of media.

Ultimately, I appreciate the attempt, but we'll have to revisit that at a later point in time and look into implementing that in a way that is better integrated and more a audio mode that can also be used as podcast.

PR #754, opened 2024-06-26
Remove Truenas SCALE From  Readme

Heavybullets8:
TrueNAS SCALE is [removing Kubernetes](https://forums.truenas.com/t/the-future-of-electric-eel-and-apps/5409).
This has caused me to move to TalosOS and redeploy my applications there instead. 

I will also be removing the site that is linked in the documentation, as all of it will be deprecated in a few months unfortunately. 

[Documentation PR](https://github.com/tubearchivist/docs/pull/44)


bbilly1:
Interesting. Well, thanks for providing the docs, at least for while it lasted. :-)

Issue #752, opened 2024-06-20
[Bug]: FileNotFoundError when fetching metadata for downloaded video.


AdamGaffney96:
### I've read the documentation

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

Synology

## Describe the bug
When attempting to download any YouTube video manually, I get the log output below and the video just remains in the queue undownloaded. I have tried restarting the containers, plus re-pulling the images also using my attached docker-compose to no success. I specify manual because my scheduled downloads still work, I wake up every day to a couple new videos successfully downloaded by the scheduled process overnight from my subscriptions. However if I try to manually add videos using the "Add to download queue" button this happens every time.

## Steps To Reproduce
- Copy a YouTube link
- Click "Add to download queue" and paste link
- Click either "Add to queue" then hit "Start download", or just use "Download now", both have the same issue

## Expected behavior
Video should download normally

### Relevant log output

```shell
20/06/2024
11:00:40
[2024-06-20 11:00:40,869: INFO/MainProcess] Task extract_download[111a0cae-0db8-4453-85fe-f532a8a0a4a0] received
20/06/2024
11:00:41
[2024-06-20 11:00:40,870: WARNING/ForkPoolWorker-5] extract_download create callback
20/06/2024
11:00:41
[2024-06-20 11:00:41,782: WARNING/ForkPoolWorker-5] HwF229U2ba8: [1/1]: add to queue
20/06/2024
11:00:46
[2024-06-20 11:00:46,882: INFO/MainProcess] Task download_pending[fe5ad2d1-ccda-4eba-85df-daf48af3bf0e] received
20/06/2024
11:00:46
[2024-06-20 11:00:46,884: WARNING/ForkPoolWorker-2] download_pending create callback
20/06/2024
11:00:46
[2024-06-20 11:00:46,928: WARNING/ForkPoolWorker-5] 111a0cae-0db8-4453-85fe-f532a8a0a4a0 success callback
20/06/2024
11:00:46
[...]

PR #757, opened 2024-06-29
workaround issue #752

Jauchi:
Hello!

Users in Issue #752 noted that getting a `403: Forbidden` error on just the audio of a video will leave some artifacts behind in the download directory, which will then cause TA to fail everytime a download of the video in question is attempted.
This PR will identify the situation and clear the download cache.

Tested it with my local installation, works perfectly for me - had several 403's, all could be resumed with this patch installed.
Now, just to clarify: TA will still get an error, but you can "try again" without restarting TA.

Thanks for the great work, keep it up!

### DIFF
--- a/tubearchivist/home/src/download/yt_dlp_base.py
+++ b/tubearchivist/home/src/download/yt_dlp_base.py
@@ -10,6 +10,7 @@ from http import cookiejar
 from io import StringIO
 
 import yt_dlp
+from home.src.ta.helper import clear_dl_cache
 from home.src.ta.settings import EnvironmentSettings
 from home.src.ta.ta_redis import RedisArchivist
 
@@ -53,6 +54,9 @@ class YtWrap:
                 print(f"{url}: failed to download with message {err}")
                 if "Temporary failure in name resolution" in str(err):
                     raise ConnectionError("lost the internet, abort!") from err
+
+                # Clear download cache directory after failures
+                clear_dl_cache(EnvironmentSettings.CACHE_DIR)
 
                 return False, str(err)
 


HarryFreeMyLand:
I was made aware of this patch on the discord and wanted to test it. but my TA doesn't seem to be reaching that part of the code that was patched to catch the exception. it seems to be bailing out a few lines above that.

below is the attached traceback 

```
Traceback (most recent call last):
  File "/root/.local/lib/python3.11/site-packages/celery/app/trace.py", line 453, in trace_task
    R = retval = fun(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/root/.local/lib/python3.11/site-packages/celery/app/trace.py", line 736, in __protected_call__
    return self.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/home/tasks.py", line 128, in download_pending
    videos_downloaded = downloader.run_queue(auto_only=auto_only)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/home/src/download/yt_dlp_handler.py", line 73, in run_queue
    continue
           ^^
  File "/app/home/src/index/video.py", line 398, in index_new_video
    video.build_json()
  File "/app/home/src/index/video.py", line 148, in build_json
    self.add_player(media_path)
  File "/app/home/src/index/video.py", line 248, in add_player
    vid_path = media_path or self.build_dl_cache_path()
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/home/src/index/video.py", line 244, in build_dl_cache_path
    raise FileNotFoundError
FileNotFoundError
```

Jauchi:
Hello!

What you posted is the _intial_ error you get. If you try to reattempt one of these videos, it'll error out instead.
My patch will just allow you to reattempt the video, it won't stop it from occuring initially.

That said, I'll see what I can do about the issue in general, maybe it's possible to just wait for a bit, when hitting the 403 error instead...

Converting to draft, I'll try to get this done when I have some time on my hands.

bbilly1:
I'm testing an approach to use the retry logic from celery: https://github.com/tubearchivist/tubearchivist/commit/1fcea860c8381d2e7a6a5ef15c5a5516152d9ba4

This should also allow continuing from cache, so if you were able to download the video but audio gave a 403, with the retry, you should just be able to retry audio only.

Jauchi:
Wow, that's way better than my quick and dirty solution.
In that case, I suppose it won't be necessary to delete the download cache and would actually break the application if that were to happen.
I actually use [Zabbix with my TA template](https://github.com/Jauchi/tubearchivist-zabbix.git) to restart failing downloads automatically, so I just wanted to click "download now" on the failed ones.

Right, so I'll just run my fork until the celery retry logic is finished - Zabbix restarts my download queue for me, so no big deal.
Closing this if you don't mind ;)

bbilly1:
> so I'll just run my fork until the celery retry logic is finished

Or even better, help with testing. I've pushed an unstable build.

Issue #721, opened 2024-05-19
[Frontend Migration]: Backend application splitup


bbilly1:
### Domain

Backend

### Description

Split the backend into more manageable and smaller django apps. App names become root path for API endpoints.

- config: as is, main django app with root settings.py
- common: All common functionality shared between apps
  - probably doesn't have any API endpoints...
- channel: all channel functionality
  - `/api/channel/`: channel list
  - `/api/channel/<id>/`: channel detail
  - `/api/channel/<id>/playlist/`: channel playlists
  - `/api/channel/<id>/shorts/`: channel shorts
  - `/api/channel/<id>/streams/`: channel live streams
  - `/api/channel/<id>/about/`: channel metadata and config
- playlist: manage all playlists
  - `/api/playlist/`: playlist list
  - `/api/playlist/<id>/`: playlist detail
- downloads: all download functionality
- settings: all application and scheduler configurations

MerlinScheurer:
The `shorts` and `streams` endpoints could be a searchparam/query `?type=short` on the `/api/channel/<id>/video/` endpoint.

bbilly1:
Completed. Tracking API changes in this gist [here](https://gist.github.com/bbilly1/7b6abc52ab689f56671bf9011879379a).

PR #768, opened 2024-07-20
DRAFT: Add Tubearchivist Frontend React dev docker setup

MerlinScheurer:
Requires:
	- Fix update note in footer: https://github.com/tubearchivist/tubearchivist/pull/780
	- Basic auth: https://github.com/tubearchivist/tubearchivist/pull/781


bbilly1:
hehe, +15k. :-)

I'm doing [#721](https://github.com/tubearchivist/tubearchivist/issues/721), maybe 50% done. That's going to move everything around. Give me some time to wrap this up first.

MerlinScheurer:
+16k now 🤣

I added Session auth for the api login, to get an [httpOnly](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#httponly) session cookie.
And calling the login endpoint twice causes it to fail.
So i had to create an ( `api/permission/user/` ) endpoint to get the isAdmin/isSuperuser to be available on every page.
If you have a better solution, feel free to suggest it 😅  ( i just don't want to save isAdmin in any browser storage )

bbilly1:
OK, let's define the scope of this PR:

- Rebase on [tubearchivist:refactor-react-migration](https://github.com/tubearchivist/tubearchivist/tree/refactor-react-migration)
- This shouldn't make any changes on the django app, that's located in `/backend`
  - If something is missing/incompatible/inconvenient in the backend, let me know, we'll deal with that outside of this PR.
- This should go to a `/frontend` folder in the root of the repo, in there you'll have `/frontend/package.json`.
- This should be able to run outside of the container in a dev environment, something like `cd frontend && npm start` should bootstrap the frontend dev environment on localhost:3000 and connect to the backend on localhost:8000
  - See CONTRIBUTING.MD with how to run the backend outside of a container
- In the Dockerfile, there should be another build step for the JS bundle, then the bundle should get loaded in the final image to /app/static so it can be served directly from nginx in the container
- This shouldn't touch any files other than the mentioned.

This PR doesn't need to ship all functionality completed. E.g. it's fine to have an application working as described above, then keep improving/extending on that from there.

MerlinScheurer:
Rebased and moved the backend changes into additional pull requests.
With this pr and the required addiational pr (https://github.com/tubearchivist/tubearchivist/pull/781), the dev env and prod deploy should work. ( only tested on windows with the backend running inside docker, since one of the packages inside the dev requirements file does not work on windows. )

PR #774, opened 2024-07-24
Feature: Enable Redis Database Option

jacaudi:
These code changes allow the end user to customize which database they connect within Redis. This is useful for users, such as myself, that run a Redis cluster and would prefer a single Redis HA cluster over running several Redis instances. For end users that have no preference, the default is set to database `0` which is the same functionally when no database is specified.

I welcome any and all feedback, thank you!!

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
     paths:
       - '**/*.py'
+  workflow_dispatch:
+    paths:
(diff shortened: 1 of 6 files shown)


bbilly1:
As you read in [CONTRIBUTING.MD](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#next-steps), we are in project freeze, we currently don't take new PRs against the old state of the code base, as that is not going to be able to be merged.

PR #764, opened 2024-07-17
add is_superuser, is_staff, and groups to /api/login

kralverde:
closes #728

PR #781, opened 2024-08-04
Add basic auth

MerlinScheurer:


bbilly1:
I'm researching this a bit. I'm not sure if basic auth is the way to go. This is fairly uncommon these days in what I read for a frontend to backend authentication.

I recall you had Session Authentication working before. What were the limitations of that? Also conveniently, we wouldn't have to open another door with another authentication schema...

MerlinScheurer:
It used the html form submit endpoint and was missing data in the response
https://github.com/tubearchivist/tubearchivist/pull/764


( but it turned out that the data is misplaced in that endpoint, we need an endpoint that can be called more the once 😅  )

Issue #725, opened 2024-05-19
[Frontend Migration]: Add API endpoint /api/channel/{channel-id}/about/


MerlinScheurer:
### Domain

Backend

## Is your feature request related to a problem? Please describe.
Data for the page `/channel/{channel-id}/about` is missing in the api endpoints.

## Describe the solution you'd like
The Api endpoint `/api/channel/{channel-id}/about/` has been implemented.

## Additional context
`/api/channel/{channel-id}/about/` ( POST )

POST body:
 ```Json
{
  "download_format": "",
  "autodelete_days": 0,
  "index_playlists": false,
  "integrate_sponsorblock": false
}
```

Possibly implemented in a module way as described in https://github.com/tubearchivist/tubearchivist/issues/721

PR #783, opened 2024-08-05
add /api/channel/{channel-id}/about/

kralverde:
closes [#725](https://github.com/tubearchivist/tubearchivist/issues/725)


bbilly1:
Sorry for the delay.

Took the liberty to simplify things, as this will now be called from an API and not though form posts, we now have proper datatypes like None and Bools to simplify life for us.

PR #787, opened 2024-08-11
Refac move sponsorblock into video endpoint

MerlinScheurer:


bbilly1:
nice, that's what we like to see. :-)

Issue #798, opened 2024-08-23
[Frontend Migration]: Reduce the calls to /api/download/aggs/


bbilly1:
### Domain

Frontend

### Description

Currently while watching the download queue progress, we should reduce the calls to refresh the download aggregations. The notification endpoint only goes to Redis, so that should be fine.

But the aggregation endpoint can be costly, particularly on a large download queue with various different channels. Ideally only refresh it after a video download completes.

PR #800, opened 2024-08-24
Fix download aggs refresh only on video finished downloading

MerlinScheurer:
https://github.com/tubearchivist/tubearchivist/issues/798


bbilly1 (approved):

bbilly1:
looks fine so far, will check things again when all put together at #798.

PR #816, opened 2024-09-23
add delay to queue.py to mitigate "Sign in to confirm you’re not a bot. This helps protect our community."

Styloy:
this adds sleep_interval to queue.py, as it was found in reindex.py. 

this helps mitigate rate-limiting errors that cause an IP to be blocked/throttled/etc when adding many videos to the download queue.

### DIFF
--- a/tubearchivist/home/src/download/queue.py
+++ b/tubearchivist/home/src/download/queue.py
@@ -6,6 +6,7 @@ Functionality:
 
 import json
 from datetime import datetime
+from time import sleep
 
 from home.src.download.subscriptions import ChannelSubscription
 from home.src.download.thumbnails import ThumbManager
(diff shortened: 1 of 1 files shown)


jcaust:
i need this!
am I to add this myself or how to ckeck/find out if it is going to be added to an update?

Styloy:
sure, if you want to test just modify the file `/app/home/src/download/queue.py` and restart the container you should have this change.

do note however that some setups will not have persistence, and you might need to make a mount, copy all of /app/* to it, move the mount to `/app/` . do note however this can break further updates.

bbilly1:
As mentioned, all code changes will need to go towards the new react refactor branch. Please rebase and add the changes there.

Styloy:
sure, I'll rebase on the react branch, apologies

PR #822, opened 2024-10-11
Add base URL support

DarkFighterLuke:
This PR adds native support for base URL and deletes the faulty workaround for nginx provided in https://gist.github.com/DarkFighterLuke/4561b6bfbf83720493dc59171c58ac36.

The implementation concerns the current UI and not the new ongoing React JS one.

Although with @bbilly1 we discussed about the convenience of the subdomain solution, it could be unreliable for many users, so I decided to open the PR anyway.


bbilly1:
As discussed, all code changes will need to go towards the new react frontend branch. New none critical additions that extend functionality will have low priority over other more pressing fixes and improvements.

PR #823, opened 2024-10-11
Show a link to video/playlist/channel on YT even if deactivated

p0358:
Just a small patch to always link to YT, even if asset is supposedly deactivated + fix of `Youtube` to `YouTube`.

Reasoning:
- be able to open the video URL anyway to
  - check if it's really down (need to see with own eyes sometimes and mourn the fact of deletion at the source)
  - check for what reason was it removed (privated, copyright claim, ToS violation, etc.)
- the video might actually be up and the link work
  - video could have been restored in the meantime and that not picked up yet (unprivatised, counter-claim filed, ToS appeal filed, etc.)
  - TA/yt-dlp could have gotten deletion status wrong, for example due to intermittent internet connectivity issues, YouTube outage or rate limiting that was interpreted as 404 etc. (it happens ¯\\_(ツ)_/¯)

I believe this small QoL improvement is worth it, as I found myself often just copying the ID from URL manually to open the link up in YT.

I was considering making "Deactivated" link non-highlighted like a regular text, "silently clickable", but eventually decided that'd be bad for accessibility and add needless complexity in CSS for no real benefit...

### DIFF
--- a/tubearchivist/home/templates/home/channel_id_about.html
+++ b/tubearchivist/home/templates/home/channel_id_about.html
@@ -45,11 +45,11 @@
         <div class="info-box-item">
             <div>
                 <p>Last refreshed: {{ channel_info.channel_last_refresh }}</p>
-                {% if channel_info.channel_active %}
-                    <p>Youtube: <a href="https://www.youtube.com/channel/{{ channel_info.channel_id }}" target="_blank">Active</a></p>
-                {% else %}
-                    <p>Youtube: Deactivated</p>
-                {% endif %}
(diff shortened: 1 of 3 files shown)


bbilly1:
As [noted](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#next-steps) we don't take any new PRs against the old code base. 

Ultimately the issue with deactivated documents is not really a frontend problem, that's results from getting blocked by YT server side and will have to be addressed at the source like that.

PR #844, opened 2024-12-06
Update yt-dlp to 2024.12.03

trick77:
The current version of yt-dlp prevents downloads, consistently returning a 403 error during the process (my installation is in a big data center). This PR addresses the issue.

**Temporary Workaround:**
To manually resolve the issue until the container image is updated, you can upgrade yt-dlp directly in the main container:
```
pip install --upgrade yt-dlp --user
```
Note: These changes will be reverted upon container restart.

### DIFF
--- a/tubearchivist/requirements.txt
+++ b/tubearchivist/requirements.txt
@@ -11,4 +11,4 @@ requests==2.32.3
 ryd-client==0.0.6
 uWSGI==2.0.28
 whitenoise==6.8.2
-yt-dlp[default]==2024.11.4
+yt-dlp[default]==2024.12.3


bbilly1:
Just updating requirements.txt is the easy part. The difficult part is to make sure the newest version is compatible with all API integrations of this project.

Datacenter IP ranges are always going to get blocked, usually VPN/Proxy will help.

trick77:
That‘s why I included the manual patching option in case it helps someone else.

LadySmith:
I can also confirm this issue with the TA yt-dlp version in my case. I haven't been able to download anything for the past 6 hours. While yt-dlp's release `2024.12.06` fixes the problem, upgrading to this version temporarily broke TA functionality in my case. 

Specifically, I temporarily had issues where the TA companion couldn't add videos to the queue, and the `start download` button for the queue didn't work, though individual download buttons remained functional. 

Consider testing thoroughly before pushing updates globally, as I've been experiencing random issues after the update. For context, I'm using a static residential IP that I've had for about 10 years.

trick77:
You may want to pin it to 2024.12.3 instead. I don't have any problems with downloading from the companion app or the download-all button using this version. I haven't tried it with the latest .6 version you are using though.
```
pip install --upgrade yt-dlp==2024.12.3 --user
```
(after restarting and thus resetting the previous modification)

As for using a proxy/VPN (initial suggestion above): I will add support for the --proxy parameter in yt-dlp in my fork. There should be no need to wrap the entire installation around a Gluetun or similar and in my case wouldn't work anyway since I'm using a combination of a reverse Docker proxy (traefik) and authentication forwarding using Authentik. 

However: most VPN providers I know have their servers in data centers (which are getting carped banned by Google), so it may not even matter but at least makes switching IP addresses for yt-dlp traffic easier. Some offer residential IPs but usually in a different pricing tier.

PR #875, opened 2025-02-09
Improve subtitle selection

### LINKED ISSUE #873: [Bug]: Unable to choose dialectical variant of downloaded subtitles
### I've read the documentation

- [x] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

Linux

## Describe the bug

Hello. I was studying how Tube Archivist handles subtitle configuration and I just realized there's a bug that leads to certain subtitles being impossible to download. The code for subtitle configuration ignores the region portion of the [BCP 47 language tag](https://en.wikipedia.org/wiki/IETF_language_tag) when storing the subtitles in a dictionary. When a video contains subtitles for more than one region, one of the regional subtitles ends up clobbering the others. This behavior is problematic.

https://github.com/tubearchivist/tubearchivist/blob/2083828fe19ba4269f4f2fcc56039d4837b19da3/backend/video/src/subtitle.py#L78-L92

## Steps To Reproduce

1. Set preferred subtitle languages to "zh" in the configuration.
2. Download the following video: https://youtu.be/tnwcLzhblKA
This video offers both Chinese (Traditional) and Chinese (Simplified) subtitles.
3. Check the subtitles downloaded by Tube Archivist.
It will be either Chinese (Traditional) or Chinese (Simplified) but not both, and you can't choose which one.
What's more, if the user were to specify "zh-TW", for example, no subtitles would be downloaded, with no indication to the user.

## Expected behavior

Since Tube Archivist ignores the region portion of the BCP 47 language tag, it would be expected for it to download both Chinese (Traditional) and Chinese (Simplified) subtitles. Although, it would be best if Tube Achivist didn't truncate the language tag as they offer important distinctions.

Although not specifically related to this bug, there's also no way to tell Tube Archivist to simply download all available subtitles, which yt-dlp does offer (you can usually tell yt-dlp "all" or specify a regex: https://github.com/yt-dlp/yt-dlp?tab=readme-ov-file#subtitle-options)

### Relevant log output

```shell
N/A
```

### Anything else?

_No response_

puigru:
This PR aims to fix #873, which prevents certain user subtitles from being downloaded, matching the behavior already in the [documentation](https://docs.tubearchivist.com/settings/application/#subtitles), while also extending its functionality to parity with [yt-dlp's --sub-langs flag](https://github.com/yt-dlp/yt-dlp?tab=readme-ov-file#subtitle-options).

The new configuration behaviors introduced in this PR work as follows:
- "all" option: The user may now specify "all" to mean all subtitle languages
- regex: The user may now use regex to specify subtitle languages (e.g. "zh.*", to download zh-Hans and zh-Hant)
- negation: The user may now prepend a dash ("-") character to their selection to exclude it (e.g. "all,-zh.*", to download all languages except all variants of Chinese)

The implementation of "orderedSet_from_options" is taken from yt-dlp:
https://github.com/yt-dlp/yt-dlp/blob/241ace4f104d50fdf7638f9203927aefcf57a1f7/yt_dlp/utils/_utils.py#L5290-L5317

### Tests
Video used: https://youtu.be/tnwcLzhblKA
Automatic caption downloading is disabled.

1. subtitle configuration is set to "all"

2. subtitle configuration is set to "zh.*"

3. subtitle configuration is set to "all,-zh.*"

### DIFF
--- a/backend/video/src/subtitle.py
+++ b/backend/video/src/subtitle.py
@@ -7,12 +7,15 @@ functionality:
 
 import json
 import os
+import re
 from datetime import datetime
+from operator import itemgetter
 
 import requests
 from common.src.env_settings import EnvironmentSettings
 from common.src.es_connect import ElasticWrap
-from common.src.helper import requests_headers
+from common.src.helper import rand_sleep, requests_headers
+from yt_dlp.utils import orderedSet_from_options
 
 
 class YoutubeSubtitle:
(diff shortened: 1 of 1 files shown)


backend/video/src/subtitle.py
```diff
@@ -7,12 +7,14 @@
...
 import os
+import re
 from datetime import datetime
+from operator import itemgetter
 
 import requests
 from common.src.env_settings import EnvironmentSettings
 from common.src.es_connect import ElasticWrap
-from common.src.helper import requests_headers
+from common.src.helper import orderedSet_from_options, requests_headers
```
bbilly1, line 17: You can import that directly from yt-dlp, so we don't need to duplicate that here:
```python
from yt_dlp.utils import orderedSet_from_options
```

backend/video/src/subtitle.py
```diff
@@ -35,85 +37,67 @@ def get_subtitles(self):
...
-                relevant_subtitles.append(user_sub)
-                continue
-
-            if self.video.config["downloads"]["subtitle_source"] == "auto":
-                auto_cap = self._get_auto_caption(lang)
-                if auto_cap:
-                    relevant_subtitles.append(auto_cap)
+        available_subtitles = self._get_all_subtitles("user")
+        if self.video.config["downloads"]["subtitle_source"] == "auto":
+            for lang, auto_cap in self._get_all_subtitles("auto"):
```
bbilly1, line 42: self._get_all_subtitles("auto") returns a dictionary, do you mean:
```python
for lang, auto_cap in self._get_all_subtitles("auto").items()
```
?

backend/video/src/subtitle.py
```diff
@@ -35,85 +37,67 @@ def get_subtitles(self):
...
 
     def download_subtitles(self, relevant_subtitles):
         """download subtitle files to archive"""
+        subtitle_list = ", ".join(map(itemgetter("lang"), relevant_subtitles))
+        print(
+            f"{self.video.youtube_id}: downloading subtitles: {subtitle_list}"
+        )
         videos_base = EnvironmentSettings.MEDIA_DIR
         indexed = []
         for subtitle in relevant_subtitles:
```
bbilly1, line 103: I tried to use the "all" method here, I got blocked quite fast. Better to slow these requests down here. you can use the `rand_sleep` function from the common/src/helper module.

bbilly1 (changes requested):
Sorry for the delay. Took some time on that, mostly looks good, added some comments here and there.

backend/common/src/helper.py
```diff
@@ -291,3 +292,59 @@ def calc_is_watched(duration: float, position: float) -> bool:
         threshold = 0.9
 
     return position >= duration * threshold
+
+
+def orderedSet_from_options(
```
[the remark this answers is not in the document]
puigru: Fixed

backend/video/src/subtitle.py
puigru: Good catch

backend/video/src/subtitle.py
puigru: Fixed

puigru:
Hey @bbilly1. Sorry, I totally missed your previous message.
I've addressed your suggestions. Let me know if there's anything else.

PR #892, opened 2025-03-11
AdminWriteOnly /appsettings/config

smantic:
In the login flow we GET /api/appsettings/config. So for non staff users to be able to login, GET requests should be available to non staff users. 

I was able to test that a non staff user was successfully able to log in after applying the patch.


bbilly1:
Nice, you got it. Thanks for fixing it!

PR #910, opened 2025-03-27
feat: dockerfile healthchecks

mmomjian:
The healthcheck can still be overridden by the user if desired but I think this is a cleaner way to integrate. No worries if you disagree. This is how most projects setup their healthchecks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,12 +77,15 @@ COPY --from=node-builder ./frontend/dist /app/static
 VOLUME /cache
 VOLUME /youtube
 
+# network
+ENV TA_PORT=8000
+
 # healthcheck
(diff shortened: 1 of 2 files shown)


lamusmaser:
@bbilly1 I think this one looks good from my side. Having the health check built-in would be helpful and I think this is maintainable. I have been testing some health checks for ES and Redis, as well, but they are still in testing phases.

bbilly1:
I appreciate the input, thanks for taking a look at that.

The problem with setting the backend port like this `ENV TA_PORT=8000`, this env var will be populated during container build time.

Yeah, healthcheck can be overwritten in the compose file, but that would need to be done explicitly by the user. If the user changes the TA_PORT, they also would need to explicitly add the healthcheck in the compose file, that is unexpected to me. If the health check is already there, it's obvious where the health check pings to and does not require the user to look at the Dockerfile implementation or at the docs to understand that.

For these reasons I prefer the current implementation.

PR #914, opened 2025-03-27
Fix URL params in embed player

masonjames:
When opening/closing embedded videos, maintain existing URL parameters including pagination state. Previously, the video player would reset page to 0 and lose other query parameters.


MerlinScheurer:
Looks good!
But you don't need to create a new URLSearchParams instance every time.
You can just modify the params  ( https://api.reactrouter.com/v7/types/react_router.SetURLSearchParams.html )

MerlinScheurer (approved):

PR #927, opened 2025-04-06
Add Ignored tab to Channel page

Jurrer:
Hello,
This change will allow to access the ignored videos from the channel page.

I find it really useful when managing a high number of channels with ignored videos. The dropdown list from downloads page is more difficult to search and does not always contain all the channels.

The button is only displayed when a channel has ignored videos


Jurrer:
Example:
image

lamusmaser:
@bbilly1 I remember us talking about this in Discord maybe two years back. It was something we didn't see a reason for at the time, but we have had more occurrences of the limitations of the ignored filter causing problems and this would solve part of those problems.
Code looks good, but up to @bbilly1 for final approval and merge.

bbilly1:
Thanks for looking into it. Looks useful.

There is a problem though when you are landing on a channel filter with ignore, you'll also need to handle removing the ignore query param. E.g.:

- from a channel page click on ignore, you land on /downloads/?channel=xxx&ignored=True
- then untoggle "Show only ignored videos"
- this should handle the refresh and show only the queued items

bbilly1:
Are you still on board here? Or shall I take over?

Jurrer:
Hi there.
I missed the message. Yes I'm still interested. 
That is a good catch, indeed the toggle does not work when we have ignored param. 
I'll post a fix today.

Jurrer:
The logic is: user clicks a `Ignored` button in the Channel page, gets redirected to `Downloads` page with the `ignored=True` and the `channeID`.
From this point user can switch this toggle, but I'm also removing the `ignored` var from URL because if user refreshes the page with it, the switch gets turned on again.

I am not sure this is the correct approach because I'm not really a dev, but it works nicely 😅 . Tested on channels with large number of ignored videos.
I'm open to feedback 😄

Issue #932, opened 2025-04-12
[Bug]: Original exception text was: 'channel_indexed'


krufab:
### I've read the documentation

- [x] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

macOs Montery

## Describe the bug
There is an error when adding a video to the download queue. It looks like the DownloadItemSerializer expects the channel_indexed field to be always present, but it seems it's not the case, at least when starting from an empty environment in development mode.

## Steps To Reproduce
1. Start from an empty environment (empty docker volumes for elastic search and redis and new sqlite database)
2. Start all the services (docker containers, backend, frontend, celery workers)
3. Log in to thea pplication via the browser
4. Go to the Download tab
5. Add a video to the download queue, either to download it immediately or to be added in the queue
6. The request is received by the backend, but an error is raised (see log output below)

This happens with all the videos. Example: https://www.youtube.com/watch?v=0J2QdDbelmY

## Expected behavior
I'd expect the video to be added without generating an error.

### Relevant log output

```shell
12/Apr/2025 03:28:07] "GET /api/download/?filter=pending HTTP/1.1" 200 176
Internal Server Error: /api/download/
Traceback (most recent call last):
  File "/Users/johnny/workspace/github/_x_/tubearchivist/backend/.venv/lib/python3.13/site-packages/rest_framework/fields.py", line 437, in get_attribute
    return get_attribute(instance, self.source_attrs)
  File "/Users/johnny/workspace/github/_x_/tubearchivist/backend/.venv/lib/python3.13/site-packages/rest_framework/fields.py", line 102, in get_attribute
    instance = instance[attr]
               ~~~~~~~~^^^^^^
KeyError: 'channel_indexed'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/johnny/workspace/github/_x_/tubearchivist/backend/.venv/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
[...]

PR #933, opened 2025-04-12
Bugfix #932: this PR resolves the error generated by the DownloadItem…

krufab:
Bugfix #932: this PR resolves the error generated by the DownloadItemSerializer as it expects to have the channel_indexed field always set.

Fixes: #932


lamusmaser:
This will need to be rebased to `tubearchivist:testing`, instead of `master`.

Issue #934, opened 2025-04-12
[Bug]: Videos not downloaded if already in the queue


krufab:
### I've read the documentation

- [x] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

Docker image

## Describe the bug
When there is a video in the download queue, and the same video is added again, but submitted with a "Download now" request from the downloads page/Add to download queue form, then the video is not downloaded immediately (and removed from the queue), but the request for that video is ignored.

It is still possible to download the video by clicking on the "Download now" button next to the video itself, but in general, when the queue contains 100+ videos, then it's hard to navigate thorough all the videos to look for the desired one. Moreover, in version 0.5.1 it is not possible to search for a video in the download queue yet.

## Steps To Reproduce
1. Add a video to the download queue
2. Request to download the same video using the "Add to download queue" form, but submitting the form with the "Download now" button.

## Expected behavior
The application should download the video even if the video is alreay in the download queue. This would be a consistent behavior, as a video not in the queue would be downloaded.

### Relevant log output

```shell
None
```

lamusmaser:
To ensure that we understand the request: 

You want to be able to utilize the option "Download Now" as an override to an item that is in the queue, rather than always having to find the video and pressing the "Download" button from the queue?

krufab:
yes, as i've 400+ videos in the queue, and i can't search for each video when i want to download it. I'd rather just add again the same url/video_id and let the app search for it, download it and remove it from the queue.

opusol:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

QuAzI:
It would be nice to fix it and to improve search in Downloads queue by name / url
In case when TubeArchivist fetches list from feed and just put video in into queue for later download or previous download attempt failed:
- user can't download video directly by link (Add to download queue -> **Download NOW**)
- browser's plugin with `auto_start: true` option doesn't work. You will see checkbox that video downloaded... but you can't find those video in TA even by name and URI

[...]

PR #935, opened 2025-04-12
Bugfix #934: Download a video from the queue

krufab:
This task addresses bug #934, allowing to download a video which is already in the queue, by issuing a new "Download now" request with the same video id. The video status will be set to 'priority' and downloaded.


lamusmaser:
@bbilly1 based on the discussion in #934, I think this is a good inclusion and creates alignment with user expectation to application action. Code changes look good to me.
To @bbilly1 for final approval and merge.

bbilly1:
I appreciate the attempt, but I think the logic could be simplified most of that functionality is already there. I also wouldn't extend this to the channel parsing, that I think is unexpected. If you add a channel and a video from that channel is ignored, that video shouldn't get added again.

But for the video itself, that makes sense if you explicitly add that again.

I'd suggest:

- In the `extrac_dl` task, pass `auto_start` to `parse_url_list`
- In `PendingList` pass the auto_start bool flag through `_add_video` if `entry["type"] == "video"`
- then `_add_video` could look something like this:

```python
def _add_video(self, url, vid_type, auto_start: bool = False):
    """add video to list"""
    if auto_start and url in set(i["youtube_id"] for i in self.all_pending):
        PendingInteract(youtube_id=url, status="priority").update_status()
        return

    if url not in self.missing_videos and url not in self.to_skip:
        self.missing_videos.append((url, vid_type))
    else:
        print(f"{url}: skipped adding already indexed video to download.")
```

That's it. Now if you add a video through the form and with "download now", that video will get downloaded with priority, no matter the previous status. Then extrac_dl already calls `download_pending` in a separate task. `videos_added` will be an empty list, that's fine, that task didn't actually add a video. Does that make sense?

krufab:
I can have a look in the next days.
In reality, in the first iteration, i didn't want to modify the signature of the function, that's why i went with the other way.

bbilly1:
Nice, that looks good. Starts to download immediately even if it's already queued.

PR #936, opened 2025-04-12
Bugfix #932: this PR resolves the error generated by the DownloadItem…

krufab:
Bugfix #932: this PR resolves the error generated by the DownloadItemSerializer as it expects to have the channel_indexed field always set.

Fixes: #932

### DIFF
--- a/backend/download/serializers.py
+++ b/backend/download/serializers.py
@@ -12,7 +12,7 @@ class DownloadItemSerializer(serializers.Serializer):
 
     auto_start = serializers.BooleanField()
     channel_id = serializers.CharField()
-    channel_indexed = serializers.BooleanField()
+    channel_indexed = serializers.BooleanField(required=False)
     channel_name = serializers.CharField()
     duration = serializers.CharField()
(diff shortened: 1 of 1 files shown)


lamusmaser:
lgtm - @bbilly1 for final approval and merge.

bbilly1:
This would be better to make sure that boolean value is always set no matter what. The serializer error just surfaces that this is handled inconsistently.

That is implemented here in `_parse_youtube_details`:

https://github.com/tubearchivist/tubearchivist/blob/testing/backend/download/src/queue.py#L330

There `channel_indexed` should be set to `False` if there aren't any channels to look up. Can you make the change? Otherwise let me know and I'll look into it.

bbilly1:
We might need to update that for existing items in the index. You can see an example of that [here](https://github.com/tubearchivist/tubearchivist/blob/testing/backend/config/management/commands/ta_startup.py#L287-L315).

So at startup this should fix all missing keys, then new items will get indexed correctly with the changes in _parse_youtube_details.

PR #938, opened 2025-04-12
feat: Reformatted the Dockerfile to allow better maintenability and f…

krufab:
This PR is to improve the Dockerfile and the .dockerignore and .gitignore file, to allow:
- better maintainability of the Dockerfile (all required libraries are sorted alphabetically)
- reduced the number of layers and improved caching, to allow faster builds
- changed the .dockerignore file to reduce ignore all unneeded files during image build (node_modules, static files)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,6 +5,13 @@
 # Docker
 .docker
 
+# Backend development
+backend/static
+backend/staticfiles
+
+# Frontend development
+frontend/node_modules
+
 # Python
 tubearchivist/__pycache__/
 tubearchivist/*/__pycache__/
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ __pycache__
 
 # django testing
 backend/static
+backend/staticfiles
 backend/.env
 
 # vscode custom conf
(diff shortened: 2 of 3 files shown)


lamusmaser:
I like this. It would reduce the workload at the top of reloading, where we get a bunch of files removed.

@bbilly1 for final approval and merge.

bbilly1:
I honestly don't really see a lot of benefits here. 

I don't see how this results in better caching, e.g. this moves the ffmpeg COPY command down in the final image, resulting in a cache miss if anything in the backend layer changes. Also `nginx.conf` is not expected to change frequently, that's why this step is above the other COPY commands. 

Removing apt/lists/* we already do in the final image, we can add that to the build layer, but as that doesn't land in the final image anyways, there is not really any benefit there.

Splitting things up with the `\` I see others do that, but for the little complexity we have here, having things in one line is fine too, imo...

For the .dockerignore and .gitignore, that's fine, I can see depending how you run and build this, this might be beneficial for your environment. I took the liberty to add them [here](https://github.com/tubearchivist/tubearchivist/commit/c5f212f520f8b24db347ef3545e67b4e84ae0464), but I'll lose this as I don't see how the Dockerfile improvements.

PR #947, opened 2025-04-22
Feat/improve development instructions

krufab:
I've improved the instructions to run the application locally for development purposes.
I've created a couple of helper files that ii found useful to have ready when testing.
I tried to keep the structure and logic as similar as the existing part.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -18,4 +18,5 @@ venv/
 assets/*
 
 # for local testing only
+dev_assets/
 testing.sh
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,56 +124,206 @@ The documentation available at [docs.tubearchivist.com](https://docs.tubearchivi
 
 ## Development Environment
 
-This codebase is set up to be developed natively outside of docker as well as in a docker container. Developing outside of a docker container can be convenient, as IDE and hot reload usually works out of the box. But testing inside of a container is still essential, as there are subtle differences, especially when working with the filesystem and networking between containers.
+This codebase is set up to be developed natively outside of docker as well as in a docker container. Developing outside
+of a docker container can be convenient, as IDE and hot reload usually works out of the box. But testing inside of a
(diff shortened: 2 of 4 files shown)


lamusmaser:
Do you have a documentation linter that is automatically adding newlines around the 120-character mark? There are a bunch of sentences that are fragmented out around that length in the `CONTRIBUTING.md` file. Otherwise, the additions look good from a cursory glance. I can do a secondary review in more depth a bit later after getting that feedback.

krufab:
Yes, i manually cut the text around 120 chars, as it's the default line width in the PyCharm IDE, and i think it's easier to read / edit Markdown texts in this way. A single \n does not introduce changes in the formatting (i.e. it does not create a new paragraph), so the final text results in normal uncut sentences.
The text I added follows this format, and I haven't touched the original text in the rest of the document.
I can shorten it to 80 or remove the \n for consistency.

lamusmaser:
Just wanted to confirm where the behavior came from. If @bbilly1 is good with it, then it can be merged as-is.

I'll review the contents a bit later for an initial review.

bbilly1:
It's a fine balance: The instructions should be tailored to devs with an average amount of experience. Focus should be what is "special" in this project and might be "unexpected". These instructions should make as little assumptions as possible. This should require as little duplication as possible, aka if something changes in the container or the codebase, the dev instructions shouldn't need to be updated too. Who is going to remember to do that?

That's why I for example tend to refer to the run.sh script instead of writing out the exact command to start the celery worker, that is just a click away. If something changes there, I don't need to remember to also update the dev instructions. 

Also I think its save to assume, that a dev interested in contributing to this project, will be able to quickly spin up an ES and redis container, as the compose file is already committed to the repo and should understand the basics of container networking, so these notes shouldn't need to elaborate on how to do that.

Also setting up an virtual environment in Python is one of the first things people learn when starting to develop Python based, so I don't think we need to explain that here. 

Does that make sense?

PR #956, opened 2025-05-10
Add video details view

MerlinScheurer:
The goal is to show additional video Metadata in a list with filters and sorting.


MerlinScheurer:
Option A:
image

MerlinScheurer:
View Style Table:
image

Also works on channel videos:

image

And playlist videos:
(Sorting hidden)
image

bbilly1:
2025-05-22_21-45-07-screenshot

Tables are always tricky for mobile friendliness. CSS Grid is one approach, because then you can convert the row better to a tile for example.

Or alternatively, maybe quicker now, just make it scroll horizontally? As we store that in the backend, so the layout will be shared across devices, if you open TA on a mobile device after you switched to table on a desktop, things won't be readable. So if the table scrolls horizontally, so you can easily switch it over to a grid again on mobile. What do you think?

bbilly1 (approved):

Issue #591, opened 2023-11-11
[Bug]: v0.4.3: "Downloads" inaccessible when using LDAP authentication


Salvoxia:
### I've read the documentation

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

Docker

## Describe the bug

I'm running tubearchivist in Docker and use LDAP for authentication.
After upgrading to v0.4.3 I noticed the "Downloads" menu item in TA was missing.
When manually navigating to ```https://tubearchivist.mydomain.com/downloads``` I was redirected to the login page with the URL ```https://tubearchivist.mydomain.com/login/?next=/downloads/```. After logging in again I ended up in a login loop with the same URL.
After disabling LDAP authentication, logging out and back in with the django credentials the "Downloads" menu item re-appeared and worked as expected.

## Steps To Reproduce
  - Set up Tubearchivist v0.4.3 to use LDAP authentication
  - Log in
  - See "Donloads" item missing from the top navigation bar
  - Manually browse to ```https://tubearchivist.mydomain.com/login/?next=/downloads/```
  - Stuck in login loop
  
## Expected behavior
"Downloads" is displayed and functioning also when using LDAP authentication

The Docker logfiles did not change from the point the container finished starting up until after reproducing the bug with ```DJANGO_DEBUG=True```.

Best Regards,
Salvoxia

### Relevant log output

```shell
Operations to perform:
  Apply all migrations: admin, auth, authtoken, contenttypes, home, sessions
Running migrations:
  No migrations to apply.
                         ....  .....
                  ...'',;:cc,. .;::;;,'...
               ..,;:cccllclc,  .:ccllllcc;,..
            ..,:cllcc:;,'.',.  ....'',;ccllc:,..
          ..;cllc:,'..                ...,:cccc:'.
         .;cccc;..                        ..,:ccc:'.
       .ckkkOkxollllllllllllc.      .,:::;.  .,cclc;
      .:0MMMMMMMMMMMMMMMMMMMX:     .cNMMMWx.   .;clc:
[...]

PR #970, opened 2025-05-25
LDAP: Multiple Authentication Backends can be configured to support permission management

joshrivers:
The core of this PR is simply adding `django.contrib.auth.backends.ModelBackend` to the AUTHENTICATION_BACKENDS array. This change has several implications:

- LDAP-sourced users *and* users from the local "Model" database can both log in
- Usernames which are available from both sources can log in using the password from either
- The TA_USERNAME/TA_PASSWORD credentials will allow an admin user to log in, and add privileges to users which have logged in using LDAP (or create user accounts in advance for LDAP users and apply privileges)

I believe this covers the core problem reported in #591, by allowing LDAP-backed installs to manage permissions during early installation phase without resorting to shelling into the container and using a REPL.

In order to allow for different requirements, I added some additional logic and a new environment variable: `TA_LOGIN_AUTH_MODE`. My thought here is that it will probably be desirable to allow for the administrator to select either one-of-many auth backends, or certain groupings of them. The boolean flags for each auth type could be used to enable combinations, but I felt like I was making a retcon change by using those and subtly changing the meaning. With the new variable, any existing install will work exactly like it did without this PR.

Additionally, anyone who sets the `TA_LOGIN_AUTH_MODE` variable will override the values of `TA_LDAP` and `TA_ENABLE_AUTH_PROXY` with an enumerated choice that specifically turns on *one* of the three supported backends, or turns on LDAP+Local Model. I'm imagining there will be more combo values available in the future as those get improved and tested.

Documentation changes which would follow from this PR:
- Adding notes on `TA_LOGIN_AUTH_MODE` to the LDAP documentation.
- Adding expository information on whether to leave `ldap_local` enabled, or just use it for the initial configuration and then switch to `ldap` (only) mode once the needed admins are promoted.
- Noting that enabling two backend requires the administrator to either remove passwords from the model backend or be OK with both the LDAP password and model password working for a given account.

Future steps:
- Not in this PR (saving it for the next one), but I'd like to add configuration which automatically promotes a list of usernames from an environment variable with staff/admin flags. I figure this will be an even better first-run experience, but it will require some custom logic.
- It should be possible to link staff/admin flags directly to LDAP groups, but my LDAP server has a bug which makes it incompatible with the django library. Even so, I'd like to set up a test harness to build that feature as well, but I think this solution is more brittle and user-environment-specific, so the first two fixes should exist first. I'd also like to make sure that the team doesn't consider it an annoying support burden before enabling it.

Let me know if there's anywhere I went the wrong direction, or any tweaks that make the changes fit in better with the desired code standards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![Tube Archivist](assets/tube-archivist-front.jpg?raw=true "Tube Archivist Banner")  
+![Tube Archivist](assets/tube-archivist-front.jpg?raw=true "Tube Archivist Banner")
 [*more screenshots and video*](SHOWCASE.MD)
 
 <div align="center">
@@ -23,13 +23,13 @@
 ------------------------
 
 ## Core functionality
-Once your YouTube video collection grows, it becomes hard to search and find a specific video. That's where Tube Archivist comes in: By indexing your video collection with metadata from YouTube, you can organize, search and enjoy your archived YouTube videos without hassle offline through a convenient web interface. This includes:  
+Once your YouTube video collection grows, it becomes hard to search and find a specific video. That's where Tube Archivist comes in: By indexing your video collection with metadata from YouTube, you can organize, search and enjoy your archived YouTube videos without hassle offline through a convenient web interface. This includes:
 * Subscribe to your favorite YouTube channels
 * Download Videos using **[yt-dlp](https://github.com/yt-dlp/yt-dlp)**
 * Index and make videos searchable
 * Play videos
 * Keep track of viewed and unviewed videos
-  
+
 ## Resources
 - [Discord](https://www.tubearchivist.com/discord): Connect with us on our Discord server.
 - [r/TubeArchivist](https://www.reddit.com/r/TubeArchivist/): Join our Subreddit.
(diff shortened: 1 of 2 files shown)


oblivioncth:
@joshrivers Hi Josh. I don't want to stomp on your potentially in-progress changes with my own, so at the very least I was hoping you could clarify the state of the changes introduced by this PR and the corresponding docs you added.

As you noted here, `TA_LOGIN_AUTH_MODE` only alters the available Django backends, but does not perform the other litany of tasks required for the complete setup of the various auth modes it touches on (e.g. username header/logout url for forward auth, ID config for LDAP, etc); however, the change to the documentation you made marked the original flags, `TA_LDAP`, `TA_ENABLE_AUTH_PROXY`, as deprecated and AFAICT encourage use of the new variable over them, giving the false impression that the new mode selection logic is feature complete and the defacto auth selection method for new installs. Yet, as aforementioned, users still very much need to set  `TA_LDAP`/`TA_ENABLE_AUTH_PROXY` in order to make use of those auth methods due to the setup that is gated by their presence.

Before possibly changing anything myself, I wanted to clarify if the approach set forth by `TA_LOGIN_AUTH_MODE` is still WIP and perhaps the doc change was potentially a bit premature? Or if not triggering the rest of the initialization code for those auth methods in `settings.py` via `TA_LOGIN_AUTH_MODE` was an oversight?

joshrivers:
@oblivioncth Thank you so much! I got focused on making that change minimal and revisiting other things later and totally missed the behavior problem you are describing.

I'm working up the next improvement to LDAP privilege assignment, and I will resolve the problem in that PR so that the behavior aligns with the changed documentation.

joshrivers:
@oblivioncth If you can, give me a PR review on #1000 

I think I've resolved the problem you name. Let me know if I've mucked anything new up too! Thanks!

PR #981, opened 2025-06-03
fixed falsely detecting the app config to be available and then failing to load with ValueError

skilletskills3:
added ipv6 expermental feature that can be enabled through env of ipv6=true
default ipv6=false


bbilly1:
Looks good for the Appsettings situation.

Let's do all yt-dlp configuration through the UI, that's why we have a fancy interface for yt-dlp after all. :-)

Maybe make a separate PR for that, as these are two very different things.

PR #983, opened 2025-06-03
gracefully handle missing appconfig by creating default settings in e…

skilletskills3:
try except to handle error gracefully and try initializing on value error.


bbilly1:
looks good, thanks for fixing it.

PR #992, opened 2025-06-09
Add option to update yt-dlp on restart

calexander3:
The majority of times that I experience an issue where a video does not download, the solution is to upgrade the version of `yt-dlp` running in the container. It would be nice if that were more convenient.

I added a new env var to have the container try to install a newer version on start. I'm not married to this approach but I thought it would be a good way to get the conversation started. Two cons to this approach are that it requires manual intervention and that there is no compatibility guarantee with the new version of `yt-dlp`. However, these are also true of my manual workflow today. Let me know what you think. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ All environment variables are explained in detail in the docs [here](https://doc
 | ELASTIC_USER | Change the default ElasticSearch user | Optional |
 | TA_LDAP | Configure TA to use LDAP Authentication | [Read more](https://docs.tubearchivist.com/configuration/ldap/) |
 | DISABLE_STATIC_AUTH | Remove authentication from media files, (Google Cast...) | [Read more](https://docs.tubearchivist.com/installation/env-vars/#disable_static_auth) |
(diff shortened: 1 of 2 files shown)


bbilly1:
I've been against this, because this is a good way to shoot yourself into the foot, yt-dlp API has changed unannounced in the passed, mostly by accident I'm sure, but still breaking things for us, in some case even resulting in a broken index.

But people seem to be determined that this is a good idea, so OK, let's do it. We'll need to make a good warning about this, so people know the risks. 

Let's implement a release update as well as a nightly update, similar to what is documented in our docs. Also make sure to specify the folder, so it updates to the same place as it was installed initially, pip will put this to a different place by default.

I'd suggest:

- `TA_AUTO_UPDATE_YTDLP=release`: To update new releases. That should result into: 
```bash
python -m pip install --target=/root/.local/bin --upgrade "yt-dlp[default]"
```

- `TA_AUTO_UPDATE_YTDLP=nightly`: To update to the nightly builds. That should result into: 
```bash
python -m pip install --target=/root/.local/bin --upgrade --pre "yt-dlp[default]"
```

Also make sure to catch an error from that command, the start up shouldn't fail if the pip command fails, if the user for example doesn't have internet.

Issue #630, opened 2023-12-10
[Feature Request]: Option to delete all channels with zero videos


ThePBone:
### I've read the documentation

- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [feature request](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#feature-request) section.

## Is your feature request related to a problem? Please describe.

After deleting a very large playlist, including all its videos, I was left with hundreds of channels that are now empty.

It would be nice to have an option to launch a clean-up operation that automatically deletes all channels that no longer have any videos linked to them. Otherwise, each channel would have to be deleted by hand.

## Describe the solution you'd like

* Add a button to the Settings/Action page to clean up empty channels (and maybe playlists too if applicable).

## Additional context
None 

### Your help is needed!

- [ ] Yes I will work on this in the next few days or weeks.

bbilly1:
This should probably be something that should be called after a delete process anyways. Along the lines of `clean_channels` that takes a list of channel IDs, to clear channels that doesn't have any videos, and don't forget doesn't have any playlists too.

Jurrer:
Hi, it would be nice to have, as empty channels quickly pile up.
Can we add it to the roadmap?

PR #994, opened 2025-06-12
Delete channels with no videos

dieperdev:
This pull request partially completes issue #630 . It adds an option under `Settings > Actions` to delete channels that have no videos published. I'm not sure if it also checks against playlist videos, shorts and livestreams but I think this is a good start.

I did add `:any` types to API requests because API responses are "untyped", but only the eslint pre-commit hook was partially skipped. All the other API endpoints called from the frontend use one function, so this should probably be done by adding another endpoint to the API (instead of sending lots of requests). I'm not familiar with the backend or ES so I did it all from the frontend.

### DIFF
--- a/frontend/src/api/actions/deleteAllEmptyChannels.ts
+++ b/frontend/src/api/actions/deleteAllEmptyChannels.ts
@@ -1,0 +1,48 @@
+import APIClient from '../../functions/APIClient';
+
+const deleteAllEmptyChannels = async () => {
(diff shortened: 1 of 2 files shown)


bbilly1:
This will need to be handled in the backend, not in the frontend. As this needs to look up playlists, if you delete a channel like that without videos but with playlists, these playlists will get deleted too, probably unexpected.

Also needs to handle download queue updating to modify the `channel_indexed` key there.

PR #995, opened 2025-06-13
Fix not found message showing as API is loading

calexander3:
When loading a channel, I noticed the channel not found message showing until the api call to get the channel info completes. This message is redundant because if the API does return a not found, the [error](https://github.com/tubearchivist/tubearchivist/blob/master/frontend/src/pages/ChannelBase.tsx#L45) from `ChannelBase` is shown instead.


.editorconfig
```diff
@@ -0,0 +1,14 @@
+# .editorconfig
```
MerlinScheurer, line 1: Not part of the changes you wanted to make, please remove.

Dockerfile
```diff
@@ -8,7 +8,7 @@ FROM node:lts-alpine AS node-builder
 COPY ./frontend /frontend
 
 WORKDIR /frontend
-RUN npm i
```
MerlinScheurer, line 11: Not part of the changes you wanted to make, please remove.

frontend/src/pages/ChannelVideo.tsx
```diff
@@ -97,104 +97,97 @@ const ChannelVideo = ({ videoType }: ChannelVideoProps) => {
     videoId,
   ]);
 
-  if (!channel) {
-    return (
-      <div className="boxed-content">
-        <br />
-        <h2>Channel {channelId} not found!</h2>
```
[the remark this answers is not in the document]
calexander3: It is not so slow that it needs that. Plus the null state of the channel doesn't necessarily equal loading.

.editorconfig
calexander3: I did intend to make this change. It is annoying to work with prettier without corresponding configuration to your editor.

Dockerfile
MerlinScheurer: Okay, so i did use `npm i` intentionally here, since `npm ci` deleted the node_module folder and re-downloads everything, slowing down the pipeline.

`If a node_modules is already present, it will be automatically removed before npm ci begins its install.` ( source: https://docs.npmjs.com/cli/v10/commands/npm-ci?v=true#description )

.editorconfig
MerlinScheurer: Prettier is setup, so that everyone can use their own editor configuration and when doing the commit a pre-commit hook will run prettier on the changed files.

calexander3: Right but if your editor is fighting every change prettier is trying to make, it is not fun. I will add this file to the git ignore instead but it is typically shared so everyone works the same way.

Dockerfile
calexander3: That's true. I will revert then. Just curious though, under what circumstances will the node modules folder ever be present when doing a docker build?

.editorconfig
MerlinScheurer: How did you setup the prettier extension in your IDE?
Because for me its, not fighting at all 🤔 

I am not completely against the editorconfig, its just an unexpected change in this pull request, given the title "Fix not found message showing as API is loading" 😅 
So if you still want to add it, please do so in a new pull request.

Dockerfile
MerlinScheurer: Local development using docker.
And maybe on the build server, but that just a guess right now.

PR #996, opened 2025-06-15
Add test notification button

calexander3:
While I have been trying to configure a webhook to get Plex to rescan my Tubearchivist when a download completes, I keep having to download an actual video to test. It has been a lot of trial and error unfortunately.

So, to speed up my configuration, I have added a test button so you can see any errors without having to actually trigger the event.

Apprise does not return these errors directly. Instead they expose out a [log capture](https://github.com/caronc/apprise/wiki/Development_LogCapture) for this purpose. It seems a bit heavy handed to me but this seems the be the way people are doing this sort of thing from Apprise.

### DIFF
--- a/frontend/src/pages/SettingsScheduling.tsx
+++ b/frontend/src/pages/SettingsScheduling.tsx
@@ -12,6 +12,9 @@ import createTaskSchedule from '../api/actions/createTaskSchedule';
 import createAppriseNotificationUrl, {
   AppriseTaskNameType,
 } from '../api/actions/createAppriseNotificationUrl';
+import testAppriseNotificationUrl, {
+  TestNotificationResponseType,
+} from '../api/actions/testAppriseNotificationUrl';
 import deleteAppriseNotificationUrl from '../api/actions/deleteAppriseNotificationUrl';
 import { ApiError, ApiResponseType } from '../functions/APIClient';
 
@@ -62,6 +65,7 @@ const SettingsScheduling = () => {
   const [downloadPendingError, setDownloadPendingError] = useState<string | null>(null);
   const [thumnailCheckError, setThumnailCheckError] = useState<string | null>(null);
   const [zipBackupError, setZipBackupError] = useState<string | null>(null);
+  const [testNotificationMessage, setTestNotificationMessage] = useState<string | null>(null);
 
   const { data: appriseNotificationData } = appriseNotification ?? {};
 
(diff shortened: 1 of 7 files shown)


frontend/src/pages/SettingsScheduling.tsx
```diff
@@ -536,49 +543,99 @@ const SettingsScheduling = () => {
...
+                }}
+              >
+                <option value="">-- select task --</option>
+                <option value="update_subscribed">Rescan your Subscriptions</option>
+                <option value="extract_download">Add to download queue</option>
+                <option value="download_pending">Downloading</option>
+                <option value="check_reindex">Reindex Documents</option>
+              </select>
+
+              <input
```
calexander3, line 570: I'd love to get this input wider if anyone has any suggestions on how to get it to fill without pushing other elements onto the next line.

bbilly1:
I appreciate trying to improve that, but I think this is a too big additional maintenance burden. If we hook into the internals of apprise like that, we run into the danger, if they change something, things will break for us as well, requiring additional maintenance on our part for that. 

I can see how testing webbooks like that can be annoying if you need to manually download a video first. But I don't expect people needing to do this very often, that's probably more a set once and forget thing. 

How about, instead of parsing apprise like that, just implement a "test" button, then log the result into the docker logs? Then we could avoid hooking into the internals of apprise like that. In the `/task/src/notify/Notifications` module you could implement a `test` method, that sends a simple "test from task_name" message? Then you can monitor the docker logs or the service you are trying to target with the notification to check if the message arrives. We wouldn't need to hook into the apprise internals, they are unlikely going to change how they handle the main interface into the apobj object there so this wouldn't add to the maintenance burden of this project?

backend/config/settings.py
```diff
@@ -329,3 +329,21 @@
...
+LOGGING = {
+    "version": 1,
+    "disable_existing_loggers": False,
+    "handlers": {
+        "console": {
+            "class": "logging.StreamHandler",
+        },
+    },
+    "loggers": {
+        "apprise": {
```
calexander3, line 343: This is needed for the Django logger to include the logs from Apprise

calexander3:
@bbilly1 Yeah that's a fair point. I have updated the pr to no longer intercept and parse the logs. It simply instructs the user to check the logs from the container to see the details.

bbilly1:
OK, looks good, took the liberty to implement the `test` method on the `/task/src/notify/Notifications` class. So we can keep the views slim, consolidate related functionality, cleaner imports, etc. 

Thanks for your addition.

PR #1002, opened 2025-06-23
Stats API calls return 0 instead of None

tnewbury:
Stats API calls returned None which meant that any other scripts/apps using the stats API endpoints had to handle None or Null.
Swapping to 0, especially in stats pages, means upstream extensions don't have to handle it.


calexander3:
I don't think this yields in the intended behavior. The responses should be objects with properties that have values of zero.

PR #1012, opened 2025-07-09
fix: Country/language code mix-up for subtitles in search examples

oblivioncth:
The parameter needs to be an ISO 639 language code, not a 3166 country code.


frontend/src/components/SearchExampleQueries.tsx
```diff
@@ -95,7 +95,7 @@ const SearchExampleQueries = () => {
               <span>full:</span> — search in video subtitles
               <ul>
                 <li>
-                  <span>lang:</span> — subtitles language (use two-letter ISO country code, same as
+                  <span>lang:</span> — subtitles language (use two-letter ISO 639 language code, same as
```
MerlinScheurer, line 98:
```suggestion
                  <span>lang:</span> — subtitles language (use two-letter ISO 639 language code,
                  same as the one from settings page)
```

oblivioncth: Thanks. Looks like wrap at 100? I'm not blind right? The linter output doesn't at all state what the issue was 😄?

oblivioncth: Ah, I took that as it was just noting my own change!

Apologizes though, I did skip proper repo setup by not looking too hard at `CONTRIBUTING.MD` and skipped doing `pre-commit install` since this was a really minimal change.

My bad.

PR #1020, opened 2025-08-04
Unify not found messaging

calexander3:
This pr cleans up some duplicate "No xxxxxs found" messaging and adds some loading indicators. The unstyled white message was coming from the inner list components and the bold green message was in each individual component that used the lists.

twoerrors


bbilly1:
Appreciate it! Thanks for the addition.

PR #1035, opened 2025-08-18
Fixes error message presented in the settings scheduling frontend when submitting an incorrect cron string

jnfbatista:
When you submit an incorrect cron (for example ``0 5``) the frontend doesn't surface the error message that the API returns, but only the default generic one.
image

I can see that a proper message is provided by the API:
image

 This fix makes it surface when the api returns the message in the error attribute and not in the message attribute.
image

I was having this issue, but after confirming I had the TA_HOST correctly defined, I went and inspect the request and saw that it was an easy fix.

You could change the API response, but I don't have enough context on the project to quickly drill down into it, but if you'd rather I go that route I can take a look into it. Let me know!

### DIFF
--- a/frontend/src/functions/APIClient.ts
+++ b/frontend/src/functions/APIClient.ts
@@ -49,7 +49,7 @@ const APIClient = async <T>(
     const data = await response.json();
     throw {
       status: response.status,
-      message: data?.message || 'An error occurred while processing the request.',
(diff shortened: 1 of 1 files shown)


bbilly1:
Hmm, yeah, we are inconsistent there. I think your approach is fine, as that will catch all custom errors raised from within the application.

PR #1060, opened 2025-09-17
embed subtitles to mp4

bmcdonough:
# Summary of embed subtitles feature
- Embedding subtitles as a feature to support Plex
- My specific use case was a Youtube channel that uses subtitles for additional details during the video.  This channel has no voice audio to be transcribed.
- The implementation adds a new configuration option `add_subtitles` that allows users to embed subtitles directly into downloaded MP4 files during the download process.

### Key Changes Made:

**Backend Configuration (`backend/appsettings/`):**
- Added `add_subtitles: bool` field to configuration schema in `src/config.py:38`
- Added corresponding serializer field in `serializers.py:47`
- Default value set to `False` for backward compatibility

**Download Handler (`backend/download/src/yt_dlp_handler.py`):**
- Enhanced `_build_obs()` method to initialize subtitle-related options:
  - `writesubtitles: False` (default)
  - `subtitleslangs: []` (default empty)
- Modified `_build_obs_postprocessors()` method to conditionally add FFmpegEmbedSubtitle processor:
  - When `add_subtitles` is enabled, adds postprocessor with `already_have_subtitle: True`
  - Sets `subtitleslangs` to configured subtitle language
  - Enables `writesubtitles` option
- Improved error handling in `_progress_hook()` method to handle missing `info_dict` gracefully

**Frontend Interface (`frontend/src/`):**
- Added new state variable `embedSubtitles` in `SettingsApplication.tsx:60`
- Added toggle control in UI with label "Embed Subtitles"
- Updated TypeScript interfaces in `loadAppsettingsConfig.ts:19` and `AppSettingsStore.ts:26`
- Enhanced help text to explain: "Embed subtitles adds the subtitles to the mp4 file."

**Bug Fixes:**
- Fixed variable name typo in `backend/channel/src/remote_query.py:116` (`query` → `queries`)
- Improved error handling for missing video info during download progress reporting
- added eslint/js to `frontend/packages.json` in order to pass pre-commit eslint

### Technical Implementation Details:

The feature leverages yt-dlp's built-in subtitle embedding capabilities through the `FFmpegEmbedSubtitle` postprocessor. When enabled:

1. Subtitles are downloaded in the configured language
2. FFmpeg embeds the subtitle streams directly into the MP4 container
3. The embedded subtitles remain accessible to media players like Plex
4. Original subtitle files are processed by the existing subtitle indexing system

This implementation follows the established pattern for other embedding features (metadata, thumbnails) and maintains consistency with the existing codebase architecture.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -15,6 +15,7 @@
         "zustand": "^5.0.7"
       },
       "devDependencies": {
+        "@eslint/js": "^9.33.0",
         "@types/react": "^19.1.9",
         "@types/react-dom": "^19.1.7",
         "@typescript-eslint/eslint-plugin": "^8.39.0",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -18,6 +18,7 @@
     "zustand": "^5.0.7"
   },
   "devDependencies": {
+    "@eslint/js": "^9.33.0",
     "@types/react": "^19.1.9",
     "@types/react-dom": "^19.1.7",
     "@typescript-eslint/eslint-plugin": "^8.39.0",
--- a/frontend/src/pages/SettingsApplication.tsx
+++ b/frontend/src/pages/SettingsApplication.tsx
@@ -549,9 +549,7 @@ const SettingsApplication = () => {
                       Indexing subtitles add the fulltext to the ES index. Not recommended on low
                       end hardware.
                     </li>
-                    <li>
-                      Embed subtitles adds the subtitles to the mp4 file.
-                    </li>
+                    <li>Embed subtitles adds the subtitles to the mp4 file.</li>
                   </ul>
                 </div>
               )}


bbilly1:
Thank you for taking a stab at this, I appreciate the addition, I think that could be helpful.

From an implementation standpoint, that would be better do add as part of the subtitle download process we already have and to use the subtitles TA generates instead of the ones from yt-dlp directly. 

We do some magic there especially stitching the autogenerated subtitles in to a more consubale format. Also consider that a user might change these settings after downloading, e.g. add a new language, then the reindex task downloads and indexes the subtitles based on the new settings. And if you implement the subtitle embedding after indexing, this change would also be represented in the mp4 file. At first glance, using ffmpeg in subprocess with the copy flag enabled looks like the right approach. Hope that makes sense? 

Then I appreciate this being a separate config option, as not all would want/need this, plus it adds IO.

lamusmaser:
As a quick note: are you (@bmcdonough) not using the Plex integration that pulls in the TA subtitles?

bmcdonough:
> As a quick note: are you (@bmcdonough) not using the Plex integration that pulls in the TA subtitles?

[tubearchivist-plex](https://github.com/tubearchivist/tubearchivist-plex)
@lamusmaser
I completely missed this!  I will try this out and let you know.
My use case as I mention above is unique.  The channel I watch has no spoken words to be transcribed, but instead inserts subtitles for additional information.  Since native plex was not picking up the TA subtitles, I thought my only option was to embed them.

lamusmaser:
> > As a quick note: are you (@bmcdonough) not using the Plex integration that pulls in the TA subtitles?
> 
> [tubearchivist-plex](https://github.com/tubearchivist/tubearchivist-plex) @lamusmaser I completely missed this! I will try this out and let you know. My use case as I mention above is unique. The channel I watch has no spoken words to be transcribed, but instead inserts subtitles for additional information. Since native plex was not picking up the TA subtitles, I thought my only option was to embed them.

Let me know if that covers the use case you are looking for. It should bring in any subtitles that are downloaded, so even subtitle inserts should come through properly.

I don't think you should cancel this if it covers the use case, but I am interested to see if this is working as intended or not.

bmcdonough:
> Next is to spend some time debugging tubearchivist-plex to identify what happened. If I identify something I will create an Issue over there, and reply back linking to it.

I identified the issue with tubearchivist-plex in how it parses the subtitle languages.  I submitted a Pull Request to handle languages with hyphens.
Pull Request:  [https://github.com/tubearchivist/tubearchivist-plex/pull/111](https://github.com/tubearchivist/tubearchivist-plex/pull/111)

bbilly1:
What's the thinking on that? Do you plan to implement that as described above or is that obsolete with the changes on the plex integration?

bmcdonough:
> What's the thinking on that? Do you plan to implement that as described above or is that obsolete with the changes on the plex integration?

@bbilly1 
1)  I reverted my branch back to `c800c27`, sorry about the changes yesterday.
2)  I use both.  The tubearchivist-plex and the embedded subtitles changes in this pull request.

The current release of tubearchivist-plex (v0.1.8) does not yet incorporate the subtitle language parsing fix.
I prefer the subtitles to be embedded so they are contained within the video file. 
This merge request is a configuration option that is disabled by default, it would not change any current usage.

PR #1093, opened 2025-11-17
Add Get Password From File

TheMDev:
- Added get password from file as per discussion with @bbilly1 in discord. This is an unoficial [docker compose](https://docs.docker.com/compose/how-tos/use-secrets/#build-secrets) convention that is also used by [ElasticSearch](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/install-elasticsearch-docker-configure). This feature allows passing in passwords as secretes.
- Took some time to clean up formatting in the README.md while adding documentation for the new `TA_PASSWORD_FILE` and `ELASTIC_PASSWORD_FILE` environment variables. If you want any of the formatting changes reverted just let me know.

### DIFF
--- a/backend/config/management/commands/ta_envcheck.py
+++ b/backend/config/management/commands/ta_envcheck.py
@@ -140,10 +140,13 @@ class Command(BaseCommand):
         env = os.environ
         for var in EXPECTED_ENV_VARS:
             if not env.get(var):
-                message = f"    🗙 expected env var {var} not set\n    {INST}"
-                self.stdout.write(self.style.ERROR(message))
-                sleep(60)
-                raise CommandError(message)
+                if not (var.endswith("_PASSWORD") and env.get(var + "_FILE")):
+                    message = (
+                        f"    🗙 expected env var {var} not set\n    {INST}"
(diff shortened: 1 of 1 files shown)


bbilly1:
Looks good.

But please split the code changes and the mardown changes into separate PRs, documentation is intendet to represent the state of the `:latest` image, [docs](https://github.com/tubearchivist/tubearchivist/blob/develop/CONTRIBUTING.md#documentation-changes).

So we can merge the code changes first, test it, without confusing people looking at the readme and not understanding why it doesn't work as documented for them. 

Then please also prep an update to the docs repo: https://docs.tubearchivist.com/installation/env-vars/

Maybe on the top something along the lines of:

"Some variables have an equivalent file based lookup. That is useful if you are using [docker secrets](https://docs.docker.com/engine/swarm/secrets/)."

then under the given env var you can be explicit like:

"This variable can also be read from a file: ..."

then under the specific

bbilly1:
Ah, and please don't modify the roadmap paragraph in the Readme. That is parsed on the server as is.

PR #1104, opened 2025-12-17
Add ``tag:`` filter to Search DSL (backend + frontend deep links)

JKamsker:
## What / Why
  Add tag-based filtering to the Search DSL and enable tag click-through from the UI to quickly find videos by tags.

  ## Changes

  ### Backend
  - `backend/common/src/searching.py`
    - Adds `tag` bucket to the parser map.
    - Treats `tag:` as implying `video` mode when no mode is provided.
    - Rejects `tag:` usage outside `video` mode.
    - Builds `query.bool.filter` for tags with OR semantics (`should` + `minimum_should_match: 1`) against `tags.keyword`.
    - Supports prefix matching via trailing `*` only; rejects unsupported wildcard patterns.
    - Rejects empty queries early.

  - `backend/common/views.py`
    - Returns `400` for missing/empty `query`.
    - Catches `ValueError` from DSL parsing/building and returns `400` with an error payload.
    - Adds OpenAPI documentation for the `query` parameter with `tag:` examples and constraints.

  - `backend/appsettings/index_mapping.json`
    - Adds a lowercase normalizer (`to_lower`) for `channel_tags.keyword` and `tags.keyword` to support consistent tag matching.

  ### Frontend
  - `frontend/src/api/loader/loadSearch.ts`
    - URL-encodes the `query` parameter (`encodeURIComponent`).

  - `frontend/src/pages/Search.tsx`
    - Reads `?query=...` from the URL and prefills the search input (enables deep-linking into searches).

  - `frontend/src/pages/Video.tsx` and `frontend/src/pages/ChannelAbout.tsx`
    - Renders tags as links to `Routes.Search?query=tag:<tag>` (URL-encoded).

  - `frontend/src/components/SearchExampleQueries.tsx`
    - Adds example `tag:` queries and brief inline docs.

  ## Tests
  - `backend/common/tests/test_src/test_searching_dsl.py`
    - Covers: tag implies video mode, OR filter query shape, tags-only searches, trailing `*` prefix queries, invalid wildcard
  rejection, and `tag:` in non-video mode rejection.

  ## How to test
  - Search examples:
    - `tag:music tag:dance`
    - `tag:music-*`
    - `video:lofi tag:music tag:dance`
  - UI:
    - Click a tag on a Video or Channel page → should open Search with `?query=tag:<tag>` and prefilled input/results.
  - Error handling:
    - `channel:linux tag:music` → `400` with “only supported in video mode”
    - `tag:*music` → `400` with “only trailing * is supported”
    - empty query → `400`

  ## Notes
  - The mapping change may require reindexing existing documents to apply the new `keyword` normalizer consistently.

### DIFF
--- a/backend/common/src/searching.py
+++ b/backend/common/src/searching.py
@@ -231,18 +231,26 @@ class QueryBuilder:
     def _build_tag_query(tag):
         """build per-tag query clause"""
         if "?" in tag:
-            raise ValueError("unsupported wildcard for tag: only trailing * is supported")
+            raise ValueError(
+                "unsupported wildcard for tag: only trailing * is supported"
+            )
 
         if "*" in tag:
-            if tag.count("*") != 1 or not tag.endswith("*") or tag.startswith("*"):
+            if (
+                tag.count("*") != 1
+                or not tag.endswith("*")
+                or tag.startswith("*")
+            ):
                 raise ValueError(
-                    "unsupported wildcard for tag: only trailing * is supported"
+                    "unsupported wildcard for tag: "
+                    "only trailing * is supported"
                 )
 
             prefix_value = tag[:-1]
             if not prefix_value:
                 raise ValueError(
-                    "unsupported wildcard for tag: only trailing * is supported"
+                    "unsupported wildcard for tag: "
+                    "only trailing * is supported"
                 )
 
             return {"prefix": {"tags.keyword": {"value": prefix_value}}}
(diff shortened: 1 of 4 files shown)


bbilly1:
Basically in `SearchParser()._get_map()` add a tag list there to the different search types. Then add it in the different subquery building methods like `QueryBuilder()._build_video()` same as the active and channel secondary key words there. 

But I understand if you don't have the time right now. Feel free to reopen if you do.

PR #1114, opened 2026-02-07
use appropriate alias creation api

camhorn:
As documented as the appropriate way to create an index: https://www.elastic.co/guide/en/elasticsearch/reference/8.19/aliases.html#add-alias

This seems to be required for ES 9.

### DIFF
--- a/backend/appsettings/src/index_setup.py
+++ b/backend/appsettings/src/index_setup.py
@@ -292,7 +292,7 @@ class ElasticIndex:
         if settings.DEBUG:
             print(f"create alias with data: {data}")
 
-        response, status_code = ElasticWrap("_alias").put(data=data)
+        response, status_code = ElasticWrap("_aliases").post(data=data)
         if status_code not in [200, 201]:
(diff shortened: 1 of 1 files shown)


bbilly1:
Interesting, that has been working fine for me though. Is that an endpoint removed in ES9?

Ultimately, ES 9 is untested, the tagged image is currently on [8.19.0](https://github.com/tubearchivist/tubearchivist/blob/develop/docker-compose.yml#L41), so still on [v8](https://docs.tubearchivist.com/installation/docker-compose/#install-elasticsearch). But yeah, eventually, we'll want to update that.

camhorn:
Endpoint doesn't work with that payload, at least. Gives an error essentially complaining about there being an array. I guess since this doesn't seem to be documented as working with ES8 (deprecated earlier? never supposed to work?), they didn't feel a need to list it as officially deprecated with ES9.

I can confirm that this change also works with ES8. Note that the linked documentation is for ES8.19.

PR #1115, opened 2026-02-08
index names now variable

camhorn:
Returned indexes are now of the form "ta_video_v2" and thus no longer exactly match, e.g., "ta_video", breaking search.

### DIFF
--- a/backend/common/src/searching.py
+++ b/backend/common/src/searching.py
@@ -31,13 +31,13 @@ class SearchForm:
         fulltext_results = []
         if search_results:
             for result in search_results:
-                if result["_index"] == "ta_video":
+                if result["_index"].startswith("ta_video"):
                     video_results.append(result)
-                elif result["_index"] == "ta_channel":
+                elif result["_index"].startswith("ta_channel"):
                     channel_results.append(result)
-                elif result["_index"] == "ta_playlist":
+                elif result["_index"].startswith("ta_playlist"):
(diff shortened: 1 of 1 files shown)


bbilly1:
How did you get a ta_video_v2? There shouldn't have been any mapping change in the last update.

But yeah, you are right, that's what you get for [duplicate](https://github.com/tubearchivist/tubearchivist/blob/develop/backend/common/src/search_processor.py#L55) that functionality, just haven't hit that yet, because my video index hasn't rebuild.

camhorn:
So, funny story with me forcing replicas, and the validation checking `number_of_replicas`. I was going to submit a change to exclude validation of that specifically, presuming good will via other fixes.

I also ran into some issues with failing to double delete prior version indexes, so there's proabably even more fallout from the versioning.

bbilly1:
Ultimately, I think aliases are the way to go. We can even extend this to a proper multi user setup, maybe some day...

I also noticed some cases where an invalid mapping change wasn't properly triggering a reindex + alias, pushed a fix for that. But luckily, we can test these things if we need to make changes to the index.

Feel free to open a separate PR to exclude `number_of_replicas` from the settings validation, that would be [here](https://github.com/tubearchivist/tubearchivist/blob/develop/backend/appsettings/src/index_setup.py#L111-L125). I think what is the only needed change to allow for multi node ES setup.

PR #1117, opened 2026-02-08
retry redis on error

camhorn:
I was seeing the worker process die with some frequency, mostly when my redis cluster switched masters, but sometimes due to transient failures. Some monitoring and/or management of the worker process is probably still called for, but this seems to address most issues.

Not sure why this isn't the celery default, to be honest.

Also experimented with redis sentinel support, since that's something celery supports, but that didn't seem especially helpful.


camhorn:
This may have actually made my stability worse. Need to look at this in more depth.

Issue #394, opened 2022-12-28
[Feature Request]: Refactor Docker image to use s6-overlay


onedr0p:
### Already implemented?

- [X] I have read through the [wiki](https://github.com/tubearchivist/tubearchivist/wiki).
- [X] I understand the [scope](https://github.com/tubearchivist/tubearchivist/wiki/FAQ) of this project and am aware of the [known limitations](https://github.com/tubearchivist/tubearchivist#known-limitations) and my idea is not already on the [roadmap](https://github.com/tubearchivist/tubearchivist#roadmap).

## Is your feature request related to a problem? Please describe.

The docker image runs as root

## Describe the solution you'd like

Since there are multiple processes in your image, it would be nice to refactor the current docker image to use [s6-overlay](https://github.com/just-containers/s6-overlay). A good example that somewhat resembles this project would be to look at [Funkwhales AIO docker image](https://dev.funkwhale.audio/funkwhale/funkwhale-docker-all-in-one). It's also a Python application and uses nginx and celery

### Your help is needed!

- [ ] Yes I can help with this feature request!

bbilly1:
I can see the advantages of that. Biggest issue will be handling permissions correctly without breaking peoples installation.

bbilly1:
After some discussion and some research, it would appear that using s6-overlay does not solve the problem that this container needs to run as root. That will need other solutions. We'll address that in a future PR.

PR #1118, opened 2026-02-09
S6 overlay

camhorn:
Pursuant to https://github.com/tubearchivist/tubearchivist/issues/394.

Keeps much of the spirit of current sequencing and grouping, with slightly enhanced parallelization. Main functional difference is that celery processes will be restarted on failure (and not just every hour for beat).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,20 +35,25 @@ ARG TARGETPLATFORM
 COPY docker_assets/ffmpeg_download.py ffmpeg_download.py
 RUN python ffmpeg_download.py $TARGETPLATFORM
 
-FROM python:3.13.11-slim-trixie AS s6-overlay
+FROM scratch AS s6-overlay-amd64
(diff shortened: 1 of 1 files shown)


camhorn:
Confirm that this doesn’t do much for or against user support by itself, although there is some particular support in s6 for running with mixed UIDs (https://github.com/just-containers/s6-overlay?tab=readme-ov-file#dropping-privileges), or alternatively, oughtn’t make supporting the `USER` directive worse. The main benefit is towards stability (e.g, restarting worker when it dies), which was my primary motivation. Much of that benefit being foundational.

Rootless nginx ought be doable, and I’d think less complex than migrating off of it, particularly given that eliminating nginx would still leave multiple processes to manage. The official solution seems to be: use tmp directories (https://github.com/nginx/docker-nginx-unprivileged/blob/61141700fbe3bb03ee7379fd0a5cba5c9c5fc885/Dockerfile-debian.template#L131 ), Largely orthogonal to process supervision, in any case.

PR #1119, opened 2026-02-09
allow disabling auto delete per channel via 0

camhorn:
As stands, I believe current behavior is to:

- Allow channels to auto delete after zero days
- Disallow zero day deletion globally
- Provide no way to disable auto deletion on a per-channel basis
- Consider negative numbers valid both globally and per channel, to surprising result

This ought normalize things such that a zero value either globally (existing behavior) or per-channel disables auto deletion, which may or may not match intent.

### DIFF
--- a/backend/download/src/yt_dlp_handler.py
+++ b/backend/download/src/yt_dlp_handler.py
@@ -317,7 +317,7 @@ class DownloadPostProcess(DownloaderBase):
         for channel_id, value in self.channel_overwrites.items():
             if "autodelete_days" in value:
                 autodelete_days = value.get("autodelete_days")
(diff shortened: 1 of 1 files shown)


bbilly1:
Are you making progress on this?

camhorn:
I confess this got personally deprioritized, as plex watch syncing seemed a ways off. Might still get to it, but no significant progress at the moment.

lamusmaser:
If it helps, Plex watch synchronization is one of the scheduled features to be included in the release planned by end of May.

bbilly1:
That's ok, feel free to take another stab at it after.

PR #1121, opened 2026-02-13
Add table for channels

komali2:
So, originally on discord, we were discussing perhaps adding the ability to see more channels sorted by certain stats as an expansion to those little lists. I went down that path a bit, but then struggled with a good UI for it that wasn't too disruptive (e.g., if I expand one table, should it expand all? If not, what to do with the negative space under the other tables?) 

So instead I thought, maybe just making the Channel List have a table view like Videos do. I then encountered the issue I believe someone already warned me about, but I didn't understand at the time: properties like a channel's total size aren't actually properties of the given document, so must be aggregated (which I've given an example of here), and aggregation can't be sorted on because it aggregates for a given page; sorting would be the equivalent of a client-side non-paginated sort. 

So, throwing this up as a draft PR since nobody asked for a table but I already did the thing so I thought I'd leave it up at least for a chat. I can go back to implementing an expansion UX for the Settings channel aggregation tables (at the bottom), or (unlikely people will go for this) look into adding channel size, video count, etc, as channel document properties - but of course that would require a new task being added as well to occasionally calculate that. 

I mean, adding client-side sorting to this table might be better than nothing? Since the returned pages are so massive anyway, even on my main instance, several TB of video, I only have 2 pages of subscribed channels. Say the word and I'll chuck that in. 

Screenshot_13-Feb_17-10-09_22330


bbilly1:
feel free to reopen if you want to give this another shot.

komali2:
I'm up for that, sorry about the delay, had a wee month log gig to sort out, but free to take a crack at this again now. Will try as you advised, calculating when download tasks finishes and adding as a property of the channel.

komali2:
Just rewrote in a new PR https://github.com/tubearchivist/tubearchivist/pull/1180

PR #1130, opened 2026-02-26
Enable --audio-multistreams flag support and add MKV container support

dakotagrvtt:
This PR attempts to enable the download and merge multi-language audio files into a supported format (.mkv) while maintaining indexing support along with ability to watch files still within the browser.

Scope of this change:
- Allow enable global use of the --audio-multistreams yt-dlp flag via a "Enable multistream audio".
image

- As a result of this feature, this also allows users to switch the default container download format to mp4 or mkv.

- Adds channel overrides with the same settings, with global, mkv, or mp4 options for container format.
image

- Add additional audio metadata on video playback page to help determine which videos contain multiple audio tracks. _May also want to add a disclaimer on MKV downloads in playback view that the additional audio tracks are not yet accessible through web player and would need to be viewed in a supported player._
image

- MKV will not allow embedded metadata at the moment. This can be implemented later.

- MKV downloads will be viewable still within the browser via **ffmpeg transcoding**. This is a big change and will require more testing and probably additional options in the UI and/or through docker. This currently works but spikes CPU temporarily and uses caching to store the transcoded file.

_Disclaimer: AI was used heavily to implement these features. All changes were limited to just the scope of implementing this feature._

### DIFF
--- a/frontend/src/pages/ChannelAbout.tsx
+++ b/frontend/src/pages/ChannelAbout.tsx
@@ -48,8 +48,8 @@ const ChannelAbout = () => {
 
   const [downloadFormat, setDownloadFormat] = useState<string | null>(null);
   const [downloadContainer, setDownloadContainer] = useState<'mp4' | 'mkv' | null>(null);
-  const [audioMultistream, setAudioMultistream] = useState<boolean | null>(null);
-  const [audioMultistreamWarning, setAudioMultistreamWarning] = useState<string | null>(null);
+  const [audioMultistreams, setAudioMultistreams] = useState<boolean | null>(null);
+  const [audioMultistreamsWarning, setAudioMultistreamsWarning] = useState<string | null>(null);
   const [audioLanguages, setAudioLanguages] = useState<string | null>(null);
   const [autoDeleteAfter, setAutoDeleteAfter] = useState<number | null>(null);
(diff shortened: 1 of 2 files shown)


bbilly1:
mkv is unlikely a good idea for this project. mp4 can have multiple video and audio tracks, quick test with yt-dlp:

```
yt-dlp \
  -f "95-1+95-2" \
  --video-multistreams \
  --audio-multistreams \
  --merge-output-format mp4 \
  'https://www.youtube.com/watch?v=9bPIaHg11mI'
```

That creates a valid mp4 file. 

Also if this needs 1000 lines of change, that definitely goes into the restriction of what PRs are welcome: https://github.com/tubearchivist/tubearchivist/blob/develop/CONTRIBUTING.md#how-to-make-a-pull-request

Additionally, if this needs a separate merge step that can't be done by yt-dlp, that also puts this out of scope.

Over all, maintainability is key for the longevity of this project, if a feature requires major rework of core functionality of this project, it's probably not a good idea to implement that here.

AI usage is OK, if the output code is of equal or better quality than an experienced dev. But it still needs to be readable for humans too. The changes here introduced on the yt_dlp_handler are not readable to me.

So if you want to pursue this: 

- this needs to be mp4 only
- implement selection in html for what language to pick
- yt-dlp only without additional cache and merge steps
- avoid doing major changes to the core modules like that

PR #1131, opened 2026-03-02
Enable --audio-multistreams flag support

dakotagrvtt:
Scope of this change:

- Allow global use of the --audio-multistreams yt-dlp flag via a "Enable multistream audio" along with channel override settings.
- Language selection once feature is enabled with a short list of common languages + a custom tag text box.
image

- Merge audio tracks to MP4 file during download processing. Will use DASH format (better metadata) first before falling back to HLS download to extract audio files (most tracks will use HLS besides the original audio track).
- Index support for audio language tags so video player UI will reflect track language names.
image

## Compatibility notes
- No DB schema migrations needed.
- Existing videos may need redownload/reindex to populate new indexed stream metadata fields (`language`, `title`) in UI (video playback page).

### DIFF
--- a/backend/appsettings/serializers.py
+++ b/backend/appsettings/serializers.py
@@ -59,7 +59,7 @@ class AppConfigDownloadsSerializer(
     extractor_lang = serializers.CharField(allow_null=True)
     integrate_ryd = serializers.BooleanField()
     integrate_sponsorblock = serializers.BooleanField()
-    audio_multistream = serializers.BooleanField()
+    audio_multistreams = serializers.BooleanField()
     audio_languages = serializers.CharField(allow_null=True)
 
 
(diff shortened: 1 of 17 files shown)


QORTEC:
@bbilly1 I would like to request that you reconsider this pull request.

> ah, I see, so no selection in TA for what language to play? So this makes this out of scope for this project.
> Maybe, some day, we will have transcoding, but unlikely for the foreseeable future. So for the time being, we should focus on what native html video players support.

If I understand your reply correctly, the reason this pull request was rejected is that currently there is no way to select alternative audio languages in TA (TubeArchivist). While this is a valid concern, it should not block this feature, as it has merit from an archival perspective. 

If TA delays implementing multi-language download support until playback support exists, content could be permanently lost in the meantime. Adding a note/disclaimer stating: `TubeArchivist only supports playback in the default language, no other languages will be available for playback` should be sufficient.

I would also like to note that TA allows users to download video files. These files can be played using alternative players that do support multiple audio tracks.

Additionally, TA advertises Jellyfin and Plex integrations. Do those integrations not support multiple audio language tracks? Wouldn't blocking this feature also limit users who wish to use TA within that ecosystem?

PR #1151, opened 2026-04-05
Refac fix layout jumps when image loadeding

MerlinScheurer:
The aspect ration css reserves the space for the image, keeping the layout from jumping.


bbilly1:
nice, that is a noticeable improvement, especially for slower connections.

PR #1154, opened 2026-04-05
feat: Send details of videos that were downloaded when sending a "download" notification

SpikedCola:
I wanted to use the "download" notification to run a script on a video finishing downloading, but the notification just includes "download 1 video(s)" with no details about which videos were downloaded.

I changed the `run_queue` function to return a list of downloaded video details. The fields `youtube_id`, `title`, `channel_name`, and `channel_id` are stored for each video downloaded, and then a json blob is sent in addition to the message.

I've been making this change manually to the last few releases, so I thought I would submit here in case anyone else finds this interesting. Hopefully it is considered to be merged so I can stop doing this myself.

### DIFF
--- a/backend/download/src/yt_dlp_handler.py
+++ b/backend/download/src/yt_dlp_handler.py
@@ -55,10 +55,12 @@ class VideoDownloader(DownloaderBase):
         self.obs = False
         self._build_obs()
 
-    def run_queue(self, auto_only=False) -> tuple[int, int]:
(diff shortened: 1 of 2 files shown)


bbilly1:
this will send JSON to every apprise notification, e.g. on your discord bot. that is undesirable for most, as that is meant to be human readable. 

Also consider that this will send an arbitrary length string while most apprise notification URLs will truncate the length depending on the service used. So this will result in invalid JSON.

Ultimately, something like that is not a good fit for apprise notifications.

PR #1155, opened 2026-04-05
feat: add stream token for mobile clients

kraetos10:
## What

  Adds short-lived, HMAC-signed streaming URLs so external video players that can't send custom `Authorization` headers (e.g.
  VLCKit, some Android players) can still stream authenticated media files.

  ## Why

  Native iOS/tvOS apps using VLCKit have a long-standing limitation: libVLC's HTTP access module has no API for custom bearer
  tokens. Confirmed by reading the official VLC-for-iOS source — they only support URL-embedded HTTP Basic auth. Without
  server-side support, apps have to work around this with fragile local HTTP proxies that inject auth headers, which breaks on
  HTTP Range seeks.

  Same problem affects any client that wants to hand a URL to a third-party media framework.

  ## How

  Adds a new API endpoint and nginx rule. Existing web UI behavior is unchanged.

  ### New endpoint: `POST /api/video/<id>/stream-token/`
  Mints a short-lived HMAC-SHA256 signature scoped to a single video and user. TTL is 5 minutes, stored in Redis so it can be
  revoked early. Authenticated as a normal API call.

  Response:
  ```json
  { "sig": "<hex>", "expires": 1712345678 }
  ```
  
Revocation: DELETE `/api/video/<id>/stream-token/`

  Deletes the sig from Redis. Clients should call this when playback stops so leaked URLs can't outlive the session.

  New internal endpoint: GET `/api/stream-auth/`

  Called by nginx auth_request when a request to /youtube/ includes ?sig=. Validates:
  - sig + expires query params are present
  - expires hasn't passed
  - Sig exists in Redis (so revocation works)
  - HMAC recomputes to the same value
  - Request URI references the signed video ID (prevents using one video's sig for another)

  Returns 200 or 403.

  nginx changes

  The /youtube/ location now conditionally routes to /api/stream-auth/ when ?sig= is present, otherwise falls back to the
  existing /api/ping/ session auth. Fully backward compatible — web UI playback is unchanged.

  Client usage

  Short videos (< 5 min):
  1. POST `/api/video/<id>/stream-token/` → get {sig, expires}
  2. Build URL: `https://host/youtube/<channel>/<video>.mp4?sig=<hex>`
  3. Hand to player
  4. On stop: DELETE `/api/video/<id>/stream-token/` with body `{"sig": "<hex>"}`

  Longer videos:
  - Renew mid-playback ~60s before expiry by calling POST again, swap the player URL, revoke old sig.

  Security

  - Long-lived API tokens stay in the Authorization header and never appear in URLs/logs
  - Each sig is tied to one specific video — can't be reused for other videos
  - 5-minute TTL + explicit revocation on stop limits exposure window if a URL leaks
  - HMAC uses SECRET_KEY so tokens can't be forged client-side
  - Uses hmac.compare_digest for constant-time comparison

  Files touched

  - backend/video/views.py — new StreamTokenView
  - backend/video/urls.py — URL pattern
  - backend/common/views.py — new StreamAuthView (no DRF auth, validates sig instead)
  - backend/common/urls.py — URL pattern
  - docker_assets/nginx.conf — conditional auth_request for /youtube/

  Testing

  Manual:
  - Mint sig, curl /youtube/.../video.mp4?sig=X&expires=Y → 206 Partial Content
  - DELETE the sig, retry same URL → 403
  - Swap sig's video_id in URL → 403
  - Tamper with sig → 403
  - Wait 5 minutes → 403 (Redis TTL expires)
  - Request /youtube/... without ?sig= in browser (logged in) → still works via session auth

  Verified from an iOS client end-to-end:
  - Video plays, native HTTP Range seeking works
  - Seek forward/back triggers new Range requests with the same sig
  - Token auto-renews mid-playback for a 30-min video
  - Closing the player revokes the sig immediately (confirmed via redis-cli KEYS 'stream_token:*')

  Breaking changes

  None. Existing /api/ping/-based auth for /youtube/ requests without a sig is preserved.

### DIFF
--- a/backend/common/urls.py
+++ b/backend/common/urls.py
@@ -30,4 +30,9 @@ urlpatterns = [
         views.HealthCheck.as_view(),
         name="api-health",
     ),
+    path(
+        "stream-auth/",
+        views.StreamAuthView.as_view(),
+        name="api-stream-auth",
+    ),
 ]
--- a/backend/common/views.py
+++ b/backend/common/views.py
@@ -1,5 +1,10 @@
 """all API views"""
 
+import hashlib
+import hmac
+import time
+from urllib.parse import parse_qs, urlparse
+
 from appsettings.src.config import ReleaseVersion
 from appsettings.src.reindex import ReindexProgress
 from common.serializers import (
@@ -18,6 +23,8 @@ from common.src.searching import SearchForm
 from common.src.ta_redis import RedisArchivist
 from common.src.watched import WatchState
 from common.views_base import AdminOnly, ApiBaseView
+from django.conf import settings
+from django.http import HttpResponse
 from drf_spectacular.utils import OpenApiResponse, extend_schema
 from rest_framework.response import Response
 from rest_framework.views import APIView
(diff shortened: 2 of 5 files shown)


bbilly1:
We already have [DISABLE_STATIC_AUTH](https://docs.tubearchivist.com/installation/env-vars/#disable_static_auth) for exactly this usecase, I think disabling auth on these read only nginx proxy requests is preferable to introducing an additional auth surface. 

Also be careful with AI generated code / text: E.g. here in the PR you say `TTL is 5 minutes` but in code it appears to be `TTL_SECONDS = 43200`.

PR #1162, opened 2026-04-20
Add audio streaming mode

MerlinScheurer:
Adds a new endpoint and frontend changes to play audio only.

Disclaimer: 
- partly ai generated.
- Backend: ai
- Frontend: ai + several manual improvements & fixes

Features:
- [x] Play audio only
- [x] Download audio only
- [x] Works with playlist autoplay

Known limitation:
- Chrome does stream the audio track, but does not show the total length of the audio track
- Firefox loads the whole audio track instantly
- Switching between video and audio while playing, does not keep timestamp.

### DIFF
--- a/backend/video/urls.py
+++ b/backend/video/urls.py
@@ -30,4 +30,9 @@ urlpatterns = [
         views.VideoSimilarView.as_view(),
         name="api-video-similar",
     ),
+    path(
+        "<slug:video_id>/stream-mp3/",
+        views.VideoStreamMp3View.as_view(),
+        name="api-video-stream-mp3",
+    ),
 ]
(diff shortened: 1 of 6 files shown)


bbilly1:
do we really need the chunking on the backend like that? As far as I can tell, we should just be able to create an `<audio>` html tag here, no? Browser should be able to handle the range header parsing? You might have to adjust nginx.conf.

Untested...

But as soon as we do streaming and junking with ffmpeg like that we increase our complexity drastically. At least my firefox is happily playing back any mp3 I drag into it without the need for any processing.

MerlinScheurer:
Thanks for the very valid feedback.
I will try a different approach.

PR #1175, opened 2026-06-03
Casting subtitles hotfix

buttertop:
Subtitles work in the web player but never render when casting — the receiver never fetches the VTT.

Two causes:

- GoogleCast.tsx uses a relative subtitle trackContentId (resolves against the receiver's origin, not the TA host). activeTrackIds is never activated and needs to be, in order for the casted device to even be signaled to send the VTT request.

- The /youtube/ nginx location maps only mp4 in its types block, so .vtt is served as application/octet-stream and isn't rendered. The /media/ block already has the text/vtt mapping.

### DIFF
--- a/docker_assets/nginx.conf
+++ b/docker_assets/nginx.conf
@@ -30,6 +30,7 @@ server {
         alias /youtube/;
         types {
             video/mp4 mp4;
+            text/vtt vtt;
         }
     }
     
--- a/frontend/src/components/GoogleCast.tsx
+++ b/frontend/src/components/GoogleCast.tsx
@@ -165,7 +165,8 @@ const GoogleCast = ({ video, setRefresh, onWatchStateChanged }: GoogleCastProps)
       for (let i = 0; i < videoSubtitles.length; i++) {
         const subtitle = new chrome.cast.media.Track(i, chrome.cast.media.TrackType.TEXT);
 
-        subtitle.trackContentId = videoSubtitles[i].media_url;
+        // subtitle.trackContentId = videoSubtitles[i].media_url;
+        subtitle.trackContentId = `${getURL()}${videoSubtitles[i].media_url}`;
         subtitle.trackContentType = 'text/vtt';
         subtitle.subtype = chrome.cast.media.TextTrackType.SUBTITLES;
         subtitle.name = videoSubtitles[i].name;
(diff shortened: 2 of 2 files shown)


bbilly1:
Where do we stand on this @MerlinScheurer ? Is this still required after the fixes I see on develop already?

MerlinScheurer:
@bbilly1 my commits did not affect subtitles.
So this is probably still required ( didn't test it )

bbilly1:
Do you have a way to test it? I dont have any hardware for this...

MerlinScheurer:
After adding `access_log /dev/stdout;` to the nginx.conf
I can see that the vtt is getting fetched from the Cast device:
`"GET /youtube/UCby13HqmxFqWEYeHUG72TQw/YWbqsBlu-uw.de.vtt HTTP/1.1" 200 83693 "-" "Mozilla/5.0 --- DeviceType/Chromecast"`

But my chrome cast ultra is not rendering any subtitles.

If there is no external way to activate or switch subtitles, i think it would be stuck to subtitle-track 0.
Otherwise it looks fine 😅 

Maybe we add a drop down with something like this:

```
  const handleSubtitleChange = useCallback((trackId: number | null) => {
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    const cast = (globalThis as any).cast;
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    const chrome = (globalThis as any).chrome;

    const castSession = cast.framework.CastContext.getInstance().getCurrentSession();
    const mediaSession = castSession?.getMediaSession?.();
    if (!mediaSession) {
      console.log('handleSubtitleChange: No active media session...');
      return;
    }

    const activeTrackIds = trackId === null ? [] : [trackId];
    const tracksInfoRequest = new chrome.cast.media.EditTracksInfoRequest(activeTrackIds);
    console.log(tracksInfoRequest);

    mediaSession.editTracksInfo(
      tracksInfoRequest,
      function () {
        setActiveSubtitleId(trackId);
      },
      function (error: unknown) {
        console.log('Error setting subtitle track', error);
      },
    );
  }, []);
```

bbilly1:
@buttertop does this need updates to the docs? We have a specific subtitle section there:

https://docs.tubearchivist.com/settings/application/#cast

Cast is supported and maintained by users like you who are interested in that, so please look at this as from a whole project perspective, not on just fixing your one thing.

Issue #1097, opened 2025-11-25
[Installation]:failed to create query: maxClauseCount is set to 3449


janolehuebner:
### Installation instructions

- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [installation help](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#installation-help) section.

### Operating System

TrueNas (docker compose)

### Your installation or setup question

Hi,
Today I noticed that my tubearchivist instance has not downloaded any videos for about 4 weeks. I changed IPs to make sure my ip is not blocked.

when I try to rescan subscriptions I get "failed to create query: maxClauseCount is set to 3449" from elasticsearch (i guess) (see relevant log output) and nothing happens...

I have no idea what changed (I have not touched the deployment except for the update to v0.5.8.
does anyone know what i can do here?

### Relevant log output

```Shell
Language set to en
INFO:     192.168.84.3:0 - "GET /api/ping/ HTTP/1.0" 200 OK
INFO:     192.168.84.3:0 - "GET /api/download/aggs/?filter=pending HTTP/1.0" 200 OK
INFO:     192.168.84.3:0 - "GET /api/download/?filter=pending HTTP/1.0" 200 OK
INFO:     192.168.84.3:0 - "GET /api/video/?sort=downloaded&order=desc HTTP/1.0" 200 OK
INFO:     192.168.84.3:0 - "GET /api/video/?sort=published&order=desc HTTP/1.0" 200 OK
INFO:     192.168.84.3:0 - "GET /api/notification/?filter=download HTTP/1.0" 200 OK
INFO:     192.168.84.3:0 - "POST /api/task/by-name/update_subscribed/ HTTP/1.0" 200 OK
INFO:     192.168.84.3:0 - "GET /api/notification/?filter=download HTTP/1.0" 200 OK
[...]

Issue #1067, opened 2025-09-29
[Bug]: Both home and  channels stopped working after upgrade to 0.5.7


gebi:
### I've read the documentation

- [x] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [ ] I'm [beta testing](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#beta-testing) and am running the latest unstable build.
- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section. I've double checked that I don't open a yt-dlp issue here.

### Operating System

Ubuntu linux with docker-compose

## Describe the bug
Both home and channels show a forever spinner instead of the videos and channels

## Steps To Reproduce
upgraded to 0.5.7 from 0.4.17

## Expected behavior
displaying video list and channels

### Relevant log output

```shell
tubearchivist    | Internal Server Error: /api/video/
tubearchivist    | Traceback (most recent call last):
tubearchivist    |   File "/root/.local/lib/python3.11/site-packages/asgiref/sync.py", line 489, in thread_handler
tubearchivist    |     raise exc_info[1]
tubearchivist    |   File "/root/.local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 42, in inner
tubearchivist    |     response = await get_response(request)
tubearchivist    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
tubearchivist    |   File "/root/.local/lib/python3.11/site-packages/asgiref/sync.py", line 489, in thread_handler
tubearchivist    |     raise exc_info[1]
tubearchivist    |   File "/root/.local/lib/python3.11/site-packages/django/core/handlers/base.py", line 253, in _get_response_async
tubearchivist    |     response = await wrapped_callback(
tubearchivist    |                ^^^^^^^^^^^^^^^^^^^^^^^
tubearchivist    |   File "/root/.local/lib/python3.11/site-packages/asgiref/sync.py", line 439, in __call__
tubearchivist    |     ret = await asyncio.shield(exec_coro)
tubearchivist    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #1063, opened 2025-09-21
[Bug]: No channels found... v0.5.7


Zev24:
### I've read the documentation

- [x] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [ ] I'm [beta testing](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#beta-testing) and am running the latest unstable build.
- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section. I've double checked that I don't open a yt-dlp issue here.

### Operating System

 Ubuntu 24.04.3 LTS

## Describe the bug
Channels tab shows no channels found
## Steps To Reproduce
Unk, I rebooted after installing OS updates and launched TA like normal. This past week the channel tab is empty (even though files are still there). No changed to the docker compose since it was first installed and working.
## Expected behavior
Many channels appear in alphabetical order and display ones DLd recently. The home tab - recent shows vids from 2024.

### Relevant log output

```shell
docker logs tubearchivist
Language set to en
[MIGRATION] Confirming v0.5.0 table layout
    ? local DB is up-to-date.
Language set to en
Operations to perform:
  Apply all migrations: admin, auth, authtoken, contenttypes, django_celery_beat, sessions, task, user
Running migrations:
  No migrations to apply.

243 static files copied to '/app/staticfiles', 1033 post-processed.
Language set to en


                         ....  .....
                  ...'',;:cc,. .;::;;,'...
               ..,;:cccllclc,  .:ccllllcc;,..
            ..,:cllcc:;,'.',.  ....'',;ccllc:,..
          ..;cllc:,'..                ...,:cccc:'.
         .;cccc;..                        ..,:ccc:'.
       .ckkkOkxollllllllllllc.      .,:::;.  .,cclc;
      .:0MMMMMMMMMMMMMMMMMMMX:     .cNMMMWx.   .;clc:
     .;lOXK0000KNMMMMX00000KO;     ;KMMMMMNl.   .;ccl:,.
     .;:c:'.....kMMMNo........    'OMMMWMMMK:    '::;;'.
   .......     .xMMMNl           .dWMMXdOMMMO'   ........
[...]

Issue #1042, opened 2025-08-28
[Bug]: Videos are not shown (HTTP 500)


justsomebody42:
### I've read the documentation

- [x] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [ ] I'm [beta testing](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#beta-testing) and am running the latest unstable build.
- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section. I've double checked that I don't open a yt-dlp issue here.

### Operating System

Linux

## Describe the bug
When TA tries to load the list of videos (either on the home page or on a subscription page), the loading indicator does not stop and the network view shows an http 500 for all /api/video requests. E.g:

- https://tubearchivist/api/video/?channel=UC6107grRI4m0o2-emgoDnAA&watch=unwatched&sort=published&order=desc&type=videos
- https://tubearchivist/api/video/?watch=unwatched&sort=published&order=desc
## Steps To Reproduce
- https://tubearchivist/api/video/

Downloading videos itself is working fine

## Expected behavior
List of videos is shown...

### Relevant log output

```shell
The TA logs does not show any ERROR message.
```

### Anything else?

I'm testing with 0.5.7, fresh installation with empty folders.

A 0.4.13 installation that I was using before was working just fine.
I tried to migrate, but after completing the migration steps without errors, the TA instance wouldn't start due to weird elasticsearch errors and thus started fresh.

bbilly1:
Not much we can do without logs. Did you activate DJANGO_DEBUG? https://docs.tubearchivist.com/installation/env-vars/#django_debug

justsomebody42:
You're absolutely right. I thought I did, but the NFS mount thought differntly...
After cleaning up (again), it's now working, sorry for the trouble!

PR #1177, opened 2026-06-07
fix: migrate channel_tags boolean to empty list on startup

boydthomson:
## Problem

Channels without tags have `channel_tags` stored as boolean `false` in Elasticsearch rather than an empty list `[]`. The strict `ListField` serializer raises `TypeError: 'bool' object is not iterable` when rendering these records, causing:

- HTTP 500 on `GET /api/channel/` and `GET /api/video/`
- Perpetual loading spinner on the Channels and Home pages
- No error visible in logs (DEBUG is off in production)

This is the root cause reported in #1042, #1063, #1067, and #1097. The videos endpoint fails because video documents embed channel metadata, so the bad `channel_tags` value propagates there too.

## Why the existing pattern doesn't work here

The migrations for artwork URL fields (`_mig_fix_channel_art_types`) use a `term: false` query to find affected documents. That works for those fields because Elasticsearch can index and query their stored values. For `channel_tags` — mapped as `keyword` — Elasticsearch does not index a boolean `false` value as a keyword term, so `term: false` matches nothing and the documents are silently skipped.

## Fix

Use `match_all` with a Painless `instanceof` check in the update script, and `ctx.op = 'noop'` to skip documents that are already correct. This is guaranteed to find and fix the affected documents regardless of how the boolean was indexed, and is safe to run repeatedly.

Two passes are made:
1. `ta_channel` index — sets `channel_tags = []` where it is not already a list
2. `ta_video` index — fixes the embedded `channel.channel_tags` copy in video documents

## Testing

Reproduced on a live instance running v0.5.10 with data going back to before v0.5.5. After applying the equivalent ES `update_by_query` calls manually:
- `GET /api/channel/` returned all 27 channels correctly
- `GET /api/video/` returned all 901 videos correctly
- UI Channels and Home pages loaded without spinning

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/config/management/commands/ta_startup.py
+++ b/backend/config/management/commands/ta_startup.py
@@ -64,6 +64,7 @@ class Command(BaseCommand):
         self._mig_fix_playlist_description()
         self._mig_fix_missing_stats()
         self._mig_fix_channel_art_types()
+        self._mig_fix_channel_tags()
         self._mig_fix_channel_description()
         self._mig_fix_video_description()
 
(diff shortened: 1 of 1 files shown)


bbilly1:
I'm pretty sure you didn't apply the migration, that is an old problem that was fixed long time ago. I dug up the commits: 

- https://github.com/tubearchivist/tubearchivist/commit/b2337669ca30a8d8f644b43178bffe4331447a6f
- https://github.com/tubearchivist/tubearchivist/commit/f4392f43fad6378bbd7b20553dfc4902045ff8c5

Ultimately, follow the update compatibility for this project as defined here: https://github.com/tubearchivist/tubearchivist#update and you should be good to go. The links you posted to the other issues are unrelated, not sure what claude hallucinated here.

PR #1180, opened 2026-06-17
Calculate channel size, persist against Channel document, and display in new Channel Table

komali2:
This PR

1. Adds new method to the `YoutubeChannel` class that calculates through aggregation the channel video count, media size, and media duration
2. Adds persistence of these fields against the ES index for a Channel
3. Invokes calculation above on video download completion, or during migration phase for any Channel without this data 
4. Adds a Table view type to the Channel view on the frontend that allows viewing these stats and sorting by them
5. Removes the Aggregator endpoint that was previously used to display Channel aggregation stats on the channel page, and replaces it with the new indexed properties on a Channel

Questions:

1. Do we want to extract channel and videos List views into a common component?
2. Do we want to persist Channel List view column sort etc into User config? If so do we want to update `SortByType` into `VideoSortByType` so it's distinct per the Video List? 

LLM usage note: Claude was used to get an overview of the codebase and recommendations on file locations for changes, basic strategy, and some implementation details. Claude was also used to generate the Table view wholesale (JSX and CSS) (just copied from the last PR), but I checked it line by line as always. All python and most typescript was input by me by "hand," with some copilot tab-completion here and there, in emacs. 

screenshot of new channel table


backend/appsettings/index_mapping.json
bbilly1, line 135: This also needs the new stats values. Channels are synced down to videos for convenient access.

backend/channel/src/index.py
```diff
@@ -168,6 +168,31 @@ def sync_to_videos(self):
...
+            },
+            "aggs": {
+                "total_items": {"value_count": {"field": "youtube_id"}},
+                "total_size": {"sum": {"field": "media_size"}},
+                "total_duration": {"sum": {"field": "player.duration"}},
+            },
+            "size": 0,
+        }
+        response, _ = ElasticWrap("ta_video/_search").get(data=data)
+        aggs = response.get("aggregations", {})
```
bbilly1, line 185: Double check what happens here on empty channel, aka channel without videos.

bbilly1 (commented):
From a quick look at this. But I think that goes into the right direction.

komali2:
Oh I forgot the aggregator on the settings page as well

backend/channel/src/index.py
komali2: This look right? 
Screenshot_19-Jun_23-20-54_10450
Screenshot_19-Jun_23-20-44_26546
Screenshot_19-Jun_23-20-32_23350

backend/appsettings/index_mapping.json
komali2: Added!

PR #1182, opened 2026-06-18
Seek to correct location on close and reopen video

komali2:
There was a bug @h-quer found wherein: 

> When watching a video with the web UI, closing that video, and clicking it again to watch without closing/refreshing the browser window, the previous watch progress gets discarded and the video plays from the initial position (since last browser refresh/page open) again.

Resetting the `videoResponse` on component teardown (unsetting `videoId`) resolves this to avoid stale cached `videoResponse` overriding the updated progress.


komali2:
Ohhh right this is setting state in a useeffect... lemme see if I can workaround that. I seem to recall some native way to unset state on component teardown. 

EDIT: resolved

komali2:
Resolved better here: https://github.com/tubearchivist/tubearchivist/commit/eeb24b2fe6b8463a3cc0fcf6be1febc86684ab25

Issue #1103, opened 2025-12-17
[Bug]: Downloading videos from multiple channels, and deleting a channel videos were downloaded from causes entire post-processing task to abort


IAmABakuAMA:
### I've read the documentation

- [x] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [ ] I'm [beta testing](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#beta-testing) and am running the latest unstable build.
- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section. I've double checked that I don't open a yt-dlp issue here.

### Operating System

Windows 11

## Describe the bug
Earlier today, I downloaded a bunch of videos from my home page. Unfortunately I also added a few videos from an AI slop channel to the queue by mistake. I didn't notice until after they were downloaded. Once they downloaded, and the queue moved on to subsequent videos from other channels, I deleted the channel.

When all videos in the queue were downloaded, and it attempted to begin playlist and comment indexing, it reached the now deleted videos from the now deleted slop channel and failed.

This is closer to enhancement than a bug, however I feel it does qualify for a bug report. Downloading hundreds of videos and not getting comments or playlist data for any of them because one video has been removed from the system is annoying, cumbersome, and I presume unintentional behaviour.

## Steps To Reproduce
- Add videos from at least 2 separate channels to the queue
- Begin download
- Once the first video is downloaded, but before the second video finishes, go to the first channel and delete it
- Observe
- Comments will not be retrieved for the 2nd channel

## Expected behavior
This error should be handled, and the removed videos gracefully skipped (similar to what was implemented for #1032)

### Relevant log output

```shell
tubearchivist         | [2025-12-17 16:24:17,197: WARNING/ForkPoolWorker-210] {"_index":"ta_channel","_id":"UCxgD63TAfkK1ZFw0vEyEbow","found":false}
tubearchivist         | [2025-12-17 16:24:17,204: WARNING/ForkPoolWorker-210] 7f916b17-fdd2-4baf-9f36-dd4244f124f9 Failed callback
[...]

PR #1183, opened 2026-06-19
Don't stop processing downloads if channel has been deleted

komali2:
Addresses https://github.com/tubearchivist/tubearchivist/issues/1103 and another potential bug by adding similar error handling used elsewhere, checking to ensure there's `json_data` on the given object before processing it, preventing an unhandled python `AttributeError`.

### DIFF
--- a/backend/download/src/yt_dlp_handler.py
+++ b/backend/download/src/yt_dlp_handler.py
@@ -426,9 +426,21 @@ class DownloadPostProcess(DownloaderBase):
             if not channel_id:
                 break
 
-            channel = YoutubeChannel(channel_id)
-            channel.get_from_es()
-            overwrites = channel.get_overwrites()
+            try:
+                channel = YoutubeChannel(channel_id)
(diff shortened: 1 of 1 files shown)


bbilly1:
I wouldn't suggest to throw an exception if you are going to catch it right away anyways. I'd treat this as loop continue, as in fetch from ES, if it's not there, simply skip it and continue with the next.

komali2:
> I wouldn't suggest to throw an exception if you are going to catch it right away anyways. I'd treat this as loop continue, as in fetch from ES, if it's not there, simply skip it and continue with the next.

Heh, that's what I originally had and then saw this pattern elsewhere so did this instead, but yeah does feel a bit much, will switch to `continue`

komali2:
@bbilly1 should we do any printing or logging in this case?

bbilly1:
yeah, let's add a print here, indicating that we are skipping that. At some point we'll need to do proper logging...

Issue #1164, opened 2026-04-28
[Bug]: audio/video bitrate units are wrong (KiB vs Kib)


w2r4:
### I've read the documentation

- [x] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [ ] I'm [beta testing](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#beta-testing) and am running the latest unstable build.
- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section. I've double checked that I don't open a yt-dlp issue here.

### Operating System

Linux using podman compose

## Describe the bug

Observed on every video. Take the video with id=HJYLROgd6IA, for instance. Its length is 15:05 == 905 seconds, and this is how its size information (--format=399+251) looks on the TA video player webpage:

<img width="422" height="130" alt="Image" src="https://github.com/user-attachments/assets/5bafb1ca-8018-4fff-b53d-014242e66677" />

File size is correct, thus:

122 MiB / 905 seconds == 138 KiB/s bitrate. Which is nowhere near 984+114 == 1098 KiB/s. Audio + video bitrate should sum up to either 138 KiB/s, or 1098 Ki**b**/s.

## Expected behavior

Use bit-sized units with bit-sized values, or use byte-sized units with byte-sized values.

### Relevant log output

```shell
come on, why is this field required?
```

MerlinScheurer:
Can you download the file from Tubearchvist via the "Download File" button below the channel on the video page and then use
 `ffprobe -v quiet -print_format json -show_streams -show_format HJYLROgd6IA.mp4`
to get the file infos and post them here?

We use ffprobe to get the bit_rate from the file.
For more information the code is: [backend\video\src\media_streams.py](https://github.com/tubearchivist/tubearchivist/blob/develop/backend/video/src/media_streams.py)

w2r4:
<img width="1513" height="973" alt="Image" src="https://github.com/user-attachments/assets/f91ae576-185b-4e4d-9735-034239bd758b" />

[...]

PR #1185, opened 2026-06-21
fix(frontend): display audio/video bitrate in correct SI units (#1164)

studiozeroseven:
## Problem

Issue #1164 — the audio/video bitrate shown on the video page and in the table view was being formatted with the byte-size helper after converting bits→bytes, then suffixed with `/s`. The result was misleading labels like `984 KiB/s` for a value that is actually `984 * 1024 / 8 = 125952 bytes/sec ≈ 1008 kbps` — a per-second *bit* rate.

For a video with combined `984 + 114 = 1098 kbps` audio+video stream selection, the page was rendering `984 KiB/s` plus `114 KiB/s` instead of the correct `984 kbps` plus `114 kbps` (or `1.1 Mbps` total). It mixed bit-units with byte-units and silently rounded the value down by ~9% in the process.

## Fix

- Add `frontend/src/functions/humanBitrate.ts` — a dedicated formatter that takes a `bitsPerSecond` value and returns a string in SI decimal units (`bps`, `kbps`, `Mbps`, `Gbps`, `Tbps`), matching the convention yt-dlp itself uses for `--format-sort bitrate`.
- Switch the bitrate cells in `VideoListItemTable.tsx` and the stream-info line in `Video.tsx` to call `humanBitrate(stream.bitrate)` directly. The `/s` suffix is gone because the unit (`kbps`) already implies per-second.
- Drop the now-unused `bitsToBytes` import from `Video.tsx`.

`bitsToBytes` is still exported in case anyone else needs byte-size conversion elsewhere — it's a generic helper, only the bitrate display was misusing it.

## How to verify

1. Open any video page (the bug reporter used id `HJYLROgd6IA`).
2. Look at the audio and video bitrate rows in the info box.
3. Before this fix: `Video: avc1 122 KiB/s` and `Audio: mp4a 14 KiB/s`.
4. After this fix: `Video: avc1 984 kbps` and `Audio: mp4a 114 kbps`.
5. Combined bitrate ≈ `1.1 Mbps`, which matches `yt-dlp -F` output for that video.

Local typecheck (`tsc --noEmit`), production build, and `eslint .` all pass on the patched tree.

### DIFF
--- a/frontend/src/components/VideoListItemTable.tsx
+++ b/frontend/src/components/VideoListItemTable.tsx
@@ -3,12 +3,12 @@ import Routes from '../configuration/routes/RouteList';
 import { VideoType } from '../pages/Home';
 import { ViewStylesType } from '../configuration/constants/ViewStyle';
 import humanFileSize from '../functions/humanFileSize';
+import humanBitrate from '../functions/humanBitrate';
 import { FileSizeUnits } from '../api/actions/updateUserConfig';
 import { useUserConfigStore } from '../stores/UserConfigStore';
 import { useVideoSelectionStore } from '../stores/VideoSelectionStore';
 import iconChecked from '/img/icon-seen.svg';
 import iconUnchecked from '/img/icon-unseen.svg';
-import bitsToBytes from '../functions/bitsToBytes';
 
 const StreamsTypeEmun = {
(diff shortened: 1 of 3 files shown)


bbilly1:

PR #1186, opened 2026-06-21
fix(pagination): clamp page_from to ES max_result_window to prevent 500 (#746)

### LINKED ISSUE #746: [Bug]: pagination can break with certain page_sizes
### I've read the documentation

- [X] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [X] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section.

### Operating System

Linux (TrueNAS k3s)

## Describe the bug

there appears to be a bug with pagination where depending on how the page_size is set (called archive view in /settings/user) could in cause too big of a request to be sent to ES.

below is the caused error, on a channel with 9981 videos and 150 view size, causing the pagination to reach a final page 67 that will return an error 500 and this output in the logs.
`{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [10050]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"ta_video","node":"6V50WepGQHS9j3s32-DcFQ","reason":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [10050]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}}],"caused_by":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [10050]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.","caused_by":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [10050]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}}},"status":400}`

lamusmaser found the specific details in a [discord thread](https://discord.com/channels/920056098122248193/1249377347816390719) his findings below:

OK, I think I figured it out.
Page 66 has the correct configurations that are expected to come through from the Pagination class: https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/home/src/index/generic.py#L79-L145
Page 67 doesn't because page_size = 150 and page_from = (9750+150 = 9900), which means that the request is going because ElasticSearch's maximum limitation, as noted https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/home/src/index/generic.py#L132. This would normally be getting handled by the max_pages calculation here: https://github.com/tubearchivist/tubearchivist/blob/master/tubearchivist/home/src/index/generic.py#L130, but we run into an issue of 10000 not dividing evenly, so any number that could go beyond the 10000 would cause this issue. The reason 24 works is because 9981/24 = 415.875; 415 * 24 = 9960, which means that the last page is 9960 through 9984, which doesn't exceed the max_pages limit by ES. 200 also works because 10000 divides evenly by it.
If you'd open up a GitHub Issue, I can link this work and put in a recommended fix for handling the page_size when it needs to be adjusted; if we decrease the page_from, then the last page will have duplicates from the previous page.
[11:13 AM]
If there were four more videos in the results, 24 also wouldn't work for the last page, since it would increment beyond from 9984 through 10001, which would cause the error to come back from ES.

### Relevant log output

```shell
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [10050]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"ta_video","node":"6V50WepGQHS9j3s32-DcFQ","reason":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [10050]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}}],"caused_by":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [10050]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.","caused_by":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [10050]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}}},"status":400}
```

### Anything else?

_No response_

studiozeroseven:
## problem

issue #746: requesting a page beyond the 10000-hit boundary in ES returns a 500 to the user. for example, page 67 of a 9981-video channel with `page_size=150` produces `from=9900 + size=150 = 10050`, which ES rejects with:

```
illegal_argument_exception: Result window is too large, from + size must be less than or equal to: [10000] but was [10050]
```

ES returns 400; the view surfaces it as 500 because the request errors out before `validate()` ever runs (validate is what knows about the 10000 cap).

## fix

`first_guess()` now clamps `page_from` to `MAX_RESULT_WINDOW - page_size` when the request would overflow, and sets `max_hits=True` so the UI can flag the cap. `validate()` uses the same constant for consistency (was the literal 10000).

## trade-off

the last page may show a few items duplicated from the previous page (we have to back the offset up to fit in the window). this is a clear improvement over the 500 the user gets today. a fuller fix would re-issue the request with a smaller `size` on the last page, but that needs the result of a first request (so two round-trips) and a UI contract change to convey the reduced `size`. happy to take that on as a follow-up if the maintainers prefer it; in the meantime this unblocks every large archive view.

## how to test

```bash
pytest backend/common/tests/test_src/test_pagination.py -v
```

covers:
- page 67 of a 9981-video channel with `page_size=150` (the reported scenario)
- the page that just fits at the boundary (`page 66`) — unchanged
- a different `page_size` (50) to guard the math
- `MAX_RESULT_WINDOW` exposed as a class constant for callers

unit test suite still green: 81/81 pass.

---

fixes #746

### DIFF
--- a/backend/common/src/index_generic.py
+++ b/backend/common/src/index_generic.py
@@ -89,6 +89,11 @@ class Pagination:
     figure out the pagination based on page size and total_hits
     """
 
+    # elasticsearch's default index.max_result_window; ES rejects any
+    # ``from + size`` request that exceeds this with a 400 (becomes a
+    # 500 to the user).  see issue #746.
+    MAX_RESULT_WINDOW = 10000
+
     def __init__(self, request):
         self.request = request
         self.page_get = False
@@ -121,12 +126,21 @@ class Pagination:
                 i for i in range(page_get - 1, page_get - 6, -1) if i > 1
             ]
             prev_pages.reverse()
+        # clamp page_from so the ES request stays within max_result_window.
+        # without this, large channel pages that cross the 10000 hit
+        # boundary (e.g. page 67 of a 9981-video channel with page_size=150)
+        # return 500.  see issue #746.
+        safe_size = self.page_size if isinstance(self.page_size, int) else 0
+        max_hits = False
+        if page_from + safe_size > self.MAX_RESULT_WINDOW:
+            page_from = self.MAX_RESULT_WINDOW - safe_size
+            max_hits = True
         pagination = {
             "page_size": self.page_size,
             "page_from": page_from,
             "prev_pages": prev_pages,
             "current_page": page_get,
-            "max_hits": False,
+            "max_hits": max_hits,
             "params": self.params,
         }
 
@@ -136,7 +150,7 @@ class Pagination:
         """validate pagination with total_hits after making api call"""
         page_get = self.page_get
         max_pages = math.ceil(total_hits / self.page_size)
-        if total_hits >= 10000:
+        if total_hits >= self.MAX_RESULT_WINDOW:
             # es returns maximal 10000 results
             self.pagination["max_hits"] = True
             max_pages = max_pages - 1
--- a/backend/common/tests/test_src/test_pagination.py
+++ b/backend/common/tests/test_src/test_pagination.py
@@ -1,0 +1,103 @@
+"""tests for Pagination class to guard against ES max_result_window overflow.
+
+regression test for issue #746: requesting a page that would push
+``from + size`` past the 10000 hit boundary returned a 500 to the user.
+the fix clamps ``page_from`` so the ES query fits in one request and
+sets ``max_hits`` so the UI can flag the cap.
+"""
+
+from unittest.mock import patch
+
+from common.src.index_generic import Pagination
+
+
+class _StubRequest:
+    """minimal request stand-in; only ``request.GET`` and ``request.user.id``
+    are used by ``Pagination.__init__``."""
+
+    def __init__(self, page):
+        self._page = page
+        self._user = type("_User", (), {"id": 1})()
+
+    @property
+    def user(self):
+        return self._user
+
+    class _GET:
+        def __init__(self, page):
+            self._page = str(page)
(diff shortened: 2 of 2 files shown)


bbilly1:

PR #1193, opened 2026-06-29
Delete empty channels/playlists

NicoBeh:
This is an attempt at implementing #630. I've implemented and tested as such that

a playlist gets deleted when
* the deletion bool is true based on the settings page
* the playlist has no videos
* the playlist is not a custom one
* the playlist is not subscribed

a channel gets deleted when
* the deletion bool is true based on the settings page
* the channel has no videos attached anymore
* the channel is not subscribed

This logic seems good to me, but if I've forgot something let me know.

## I'm a human

- [x] I confirm that I'm a human opening this PR.


bbilly1:
Thanks for taking a look at this. I think targeting only unsubscribed channels/playlist is the right call here.

I couldn't quite follow the logic around deleting channels with playlists. Consider, some channels are only indexed because of the playlist. So even if the channel does not have videos, it might still have playlists with videos.

Also, please prepare the docs with the new features.

Issue #1199, opened 2026-07-27
[Bug]: Issue with subtitle_key output


pendletong:
### I've read the documentation

- [x] I'm running the latest version of Tube Archivist and have read the [release notes](https://github.com/tubearchivist/tubearchivist/releases/latest).
- [x] I'm [beta testing](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#beta-testing) and am running the latest unstable build.
- [x] I have read the [how to open an issue](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#how-to-open-an-issue) guide, particularly the [bug report](https://github.com/tubearchivist/tubearchivist/blob/master/CONTRIBUTING.md#bug-report) section. I've double checked that I don't open a yt-dlp issue here.

### Operating System

Docker on MacOS

## Describe the bug
If there is an issue with the video.check_subtitles() call the line print(f"{subtitle_key} skip empty subtitle') throws an error because subtitle_key hasn't been initialised

## Steps To Reproduce
The video I was downloading was https://www.youtube.com/watch?v=GJdyoJcZsuo

## Expected behavior
As the subtitle is empty, presumably it should just log skip empty subtitle.

### Relevant log output

```shell
[2026-07-27 08:32:39,002: ERROR/ForkPoolWorker-1] Task download_pending[5b7f7f27-8546-42b1-8b1a-00d5706a6082] raised unexpected: UnboundLocalError("cannot access local variable 'subtitle_key' where it is not associated with a value")
Traceback (most recent call last):
  File "/root/.local/lib/python3.13/site-packages/celery/app/trace.py", line 585, in trace_task
    R = retval = fun(*args, **kwargs)
                 ~~~^^^^^^^^^^^^^^^^^
  File "/root/.local/lib/python3.13/site-packages/celery/app/trace.py", line 858, in __protected_call__
    return self.run(*args, **kwargs)
           ~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/app/task/tasks.py", line 136, in download_pending
    downloaded, failed = downloader.run_queue(auto_only=auto_only)
                         ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/app/download/src/yt_dlp_handler.py", line 83, in run_queue
    vid_dict = index_new_video(youtube_id, video_type=video_type)
  File "/app/video/src/index.py", line 530, in index_new_video
[...]

PR #1200, opened 2026-07-27
Remove subtitle_key from log

pendletong:
subtitle_key is not available to log if there is no response text
This fixes the visible issue with #1199

Check the box, if you are a human.

## I'm a human

- [x ] I confirm that I'm a human opening this PR.

### DIFF
--- a/backend/video/src/subtitle.py
+++ b/backend/video/src/subtitle.py
@@ -193,7 +193,7 @@ class YoutubeSubtitle:
             return None
 
         if not response.text:
-            print(f"{subtitle_key}: skip empty subtitle")
+            print("skip empty subtitle")
             rand_sleep(self.video.config)
             return None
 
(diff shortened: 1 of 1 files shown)


bbilly1:
the right fix here is to move `subtitle_key = f"{self.video.youtube_id}-{lang}"` out of the conditional, so the log can still indicate what has failed. Something like this: 

```diff
diff --git a/backend/video/src/subtitle.py b/backend/video/src/subtitle.py
index 1fdd0919..8615a34e 100644
--- a/backend/video/src/subtitle.py
+++ b/backend/video/src/subtitle.py
@@ -185,8 +185,8 @@ class YoutubeSubtitle:
 
         response = requests.get(url, **request_kwargs)
 
+        subtitle_key = f"{self.video.youtube_id}-{lang}"
         if not response.ok:
-            subtitle_key = f"{self.video.youtube_id}-{lang}"
             print(f"{subtitle_key}: failed to download subtitle")
             print(response.text)
             rand_sleep(self.video.config)
```

pendletong:
Yep, that makes sense. I have corrected my initial attempt.

PR #1201, opened 2026-07-29
feat(ui): add missing units to download speed

H4ckerxx44:
makes it more clear in what kind of unit the download speed is described as, some applications would limit in MiB/s, some in KiB/s, we (or yt-dlp) chose B/s

## I'm a human

- [x] I confirm that I'm a human opening this PR.

### DIFF
--- a/frontend/src/pages/SettingsApplication.tsx
+++ b/frontend/src/pages/SettingsApplication.tsx
@@ -320,7 +320,7 @@ const SettingsApplication = () => {
               )}
               <div className="settings-box-wrapper">
                 <div>
-                  <p>Download Speed limit</p>
+                  <p>Download Speed limit (B/s)</p>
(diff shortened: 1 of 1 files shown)


lukadd16:
@bbilly1 I don't think bytes/sec is an accurate label either. Correct me if I'm wrong, but these lines would indicate to me that TA is expecting the number to be in kibibytes/sec (1 KiB = 1024 bytes), no?

https://github.com/tubearchivist/tubearchivist/blob/985311d59e8ea2451feb689ff200c33c40e54e25/backend/download/src/yt_dlp_handler.py#L177-L184

bbilly1:
hehe, well, I see people care about that. :-) 

Feel free to proceed with another PR. We do one final PR on this, after that, it is what it is.

